Issue #419, opened 2021-04-07
Remove SNAP Banked Months From Active Scripts and Functions


IlseFerris:
SNAP Banked Months are not available as we are in a federal ABAWD waiver period, and all SNAP benefits are not time-limited. The current federal waiver is through 06/30/21, and is assumed to extend for at least another year beyond that. The following scripts and functions will need to be reviewed/updated for SNAP Banked Months:

- [x] NOTES - ABAWD WAIVED APPROVAL
- [x] NOTES - ABAWD TRACKING RECORD
- [x] NOTES - CLOSED PROGRAMS
- [x] NOTICES - SNAP E & T LETTER
- [x] ACTIONS - ABAWD FIATER
- [x] NOTES - CAF (function update_wreg_and_abawd_notes)
- [x] FUNCLIB - autofill_editbox_from_MAXIS (panel_read_from = "WREG" portion)

IlseFerris:
No issues reported in past week. Closing issue.

PR #439, opened 2021-04-20
Snap banked months scrub

IlseFerris:
Updates to scripts and functions from issue #419. Initial tested, but I'd love some fresh eyes in case I miss anything.


C-Love:
@IlseFerris In the CAF script there is a dialog that allows for manual entry of the Banked Months.
image
Do you want to leave that as is? With the changes to the rest of the script anything entered there would not end up on the main dialog or in the case notes.

C-Love:
Another question - Do we need the ABAWD FIATer right now?

C-Love:
Final question - are we no longer concerned with sending 'ABAWD FS RULES' - was that specific to BM?

Also - all questions can be addressed at the meeting.

IlseFerris:
@C-Love I updated the CAF script to remove the banked months field. 
All other questions we addressed yesterday. Can you take a quick look at my latest commit?

C-Love (approved):
Awesome. Looks good. And while you have commented most of it out... can we hope that these never come back in this way.

Issue #432, opened 2021-04-14
Add Appt Notices to Review Report 


C-Love:
We need to move the functionality to send the appointment notices from the On Demand script into the Review Report script. 

The determination of which cases need a notice should be able to be determined from the details on the Review Report already - we should not need to investigate in MAXIS because the programs, renewal type and interview requirement is determined by the Review Report script.

C-Love:
These worked wonderfully! All done.

PR #447, opened 2021-04-23
#432 - Adding Appointment Notices to Review Report

C-Love:
I was able to run this successfully and send out notices and capture the detail on the spreadsheet - so all the good things we want. 

I had two errors in the run;
1. The very las statistics didn't work because the automation around statistics uses the 'Table1' name assuming the information was in a table. It was not but this was fine. We lost no information and I was able to add it in manually after creating the table. 
2. I messed up and the CASE:NOTEs may indicate that information was sent to an AREP or SWKR when those aren't things. Basically I had the blanking of variables out of the loop. Without a way to small batch test with this functionality - this month is just going to be a bit funky. It is fixed now and should work with no issue next time.

I am confident the second issues is fixed with this. 

The first issue, I can keep this issue open and add the automation of the table creation to the initial report functionality when I work on that next week IF we are concerned. It is not an inhibiting error at all if we miss it for this functionality.


IlseFerris (approved):
Looks good to me!

PR #463, opened 2021-05-14
Avs dail scrubber

IlseFerris:
Added DAIL Scrubber to run NOTES - AVS. Updated NOTES - AVS to support reading case number and defaulting initial options based on the type of the TIKL.


C-Love (approved):
Looks good - nice easy add - high impact - low lines of code. @IlseFerris for the win.

Issue #506, opened 2021-06-24
Add GRH Support to NOTICES - PA Verif Request


C-Love:
Adding functionality for GRH benefits. 
 - Sending WCOMs
 - Sending MEMOs

C-Love:
GRH is now available in PA Verif Request.

PR #535, opened 2021-07-20
#506 NOTICES - PA Verif add GRH Functionality

C-Love:
The functionality should work well here. 

I have questions about the functionality where we figure out the building of the MEMOs. 
Can we discuss?


IlseFerris (changes requested):
GRH/Housing testing worked as intended, woo hoo!
One little thing that needs to be updated. At the end of the MEMO, the text says "GRH Food Total for (time period)". That should probably not say food :)

C-Love:
AH! 
Hahaha! 

Yeah - probably not food. I will update and then pull in.

Issue #557, opened 2021-08-12
FUNCTION - Resources Notifier - error/err msg when selecting ONLY Virtual Dropbox


MiKaylaH22:
Script name - ACTIONS - INTERVIEW.vbs was run on Case #XXX with a runtime of 1596.969 seconds. 
Information: I tried sending the email/virtual dropbox info to clt and the script kept giving me an error. This is the first time I tried sending just that without any other resources, it's always worked with multiple selections.

The error said something like “Select Resource or Check None” and even though I selected the email/virtual dropbox option it kept popping up that message. I gave the clt the info verbally while trying several times so finally just hit none so that we could move on.

C-Love:
This is an issue with the Function for the resources notifier and will need to be resolved in the FuncLib.

C-Love:
This is working now. Bug squashed

Issue #542, opened 2021-07-27
Update Resources Information with WIC detail


C-Love:
Add WIC referral information to the functionality for resources.

C-Love:
This information is added.

Issue #346, opened 2021-02-16
Update the resources notifier - shelter information


C-Love:
There was a hot topic about shelter information - https://hennepin.sharepoint.com/teams/hs-economic-supports-hub/SitePages/How-do-I-connect-Hennepin-County-residents-experiencing-homelessness-to-Shelter-.aspx

Update he script with this detail.

C-Love:
This information has been added.

PR #576, opened 2021-08-27
#557, #542 & #346 FUNCTIONS LIBRARY - Resources Notifier function update

C-Love:
Updates to the Resources Notifier function so that it updates all the places that the resources are called. 

ERR Msg handling to correctly determine if something was selected or not.
Added Shelter informaiton
Added WIC information

Ready for review and pulling in.


IlseFerris (changes requested):
NOTICES - RESOURCE NOTIFIER: Totally worked. Added shelter, adult and family, information and WIC information. 
NOTES - INTERVIEW: Was unable to pass 'check here is not resources are needed'. I was able to send shelter, WIC and other resources. Sending email with screen shots.

C-Love:
Updates made for the Interview run - and any run that fit INTO another script run - to be able to select none and make sure the 'Case Number' text is gone.

IlseFerris (approved):

Issue #548, opened 2021-08-04
Update the ADD WCOM for Expedited to accommodate ALL postponed verifs


C-Love:
Current ADD WCOM functionality for Expedited with Postponed Verifications is only for WREG postponed verifs. Update this checkbox to be for ALL postponed verifs and then change the functionality to have specific handling for postponed WREG verifs and then other postponed verifs. 

This is suggested verbiage for WCOM from Mandora

Create a WCOM verbiage to go with the approved expedited SNAP notice once the case is approved for expedited.

You are getting SNAP right away because you meet the criteria for expedited SNAP.  You still need to provide the following postponed verification(s) that is needed to continue your SNAP eligibility.  We will mail out a separate verification request form for the requested verification(s).

This is connected to support for Expedited being built in:
#363 - DAIL Update - add MEMO to EXP SNAP closing DAIL
and 
#208 - Expedited Approval at Determination

C-Love:
Associated with QI Project Board - Postponed EXP SNAP Notice Verbiage in Project Board

C-Love:
This works. Has it been used? ...

PR #579, opened 2021-09-01
#548 NOTICES - Add WCOM - WCOM for ALL postponed verifs

C-Love:


IlseFerris (approved):
All mandatory field handling is working. SPEC/WCOM text correct and formatted correctly. CASE/NOTE clear and informative.

Issue #527, opened 2021-07-14
Function add sort_date to FuncLib


IlseFerris:
Add to Funclib, remove from individual scripts. 

- [x] FuncLib add
- [x] Remove from misc/hss-maxis-facility-report.vbs
- [x] Remove from misc/hss-service-authorization-update.vbs
- [x] Remove from notices/pa-verif-request.vbs
- [x] Remove from actions/earned-income-budgeting.vbs

C-Love:
These are all still working - function is now global.

PR #585, opened 2021-09-07
#527 - Move sort_dates function to FuncLib

C-Love:
Removing 'sort_dates' from individual scripts and moved it to the Functions library. Then we remove it from the 4 scripts it is already locally defined in.

HSS scripts to be tested by @IlseFerris . 

Added better commenting to the function in MASTER FUNCTIONS LIBRARY


IlseFerris (approved):
I review both the Housing Supports scripts in the MICS folder. Function and script working as anticipated. I didn't test the other scripts, assuming you got those.

C-Love:
I did get those!

Issue #575, opened 2021-08-26
PA - Verif Request - Ignore 'Cancelled' notices 


C-Love:
We should not 'resend' a notice that was cancelled and the  script should ignore these when creating the dialog/list.

Currently these just fail to resend and there is no information.

C-Love:
Appears to be working - closing.

Issue #565, opened 2021-08-18
PA-VERIF create a loop to ensure the user is not over on page limit


MiKaylaH22:

C-Love:
Also add a check to see if the date selected is past the display date (the issuances have been archived) - loop back to the dialog if this message is hit and have the dates to be changed.

C-Love:
Appears to be working - closing.

Issue #508, opened 2021-06-24
Add Export to Word Functionality to NOTICES - PA Verif Request


C-Love:
Create option for In Person or Requesting Fax to have the script generate a Word Document of each WCOM or MEMO that was generated/resent.

C-Love:
Appears to be working - closing.

Issue #507, opened 2021-06-24
Add DWP Support to NOTICES - PA Verif Request


C-Love:
Adding functionality for DWP benefits. 
 - Sending WCOMs
 - Sending MEMOs

C-Love:
Appears to be working - closing.

PR #600, opened 2021-09-21
#507, #508, #565, #575 Updates to NOTICES - PA Verif Request 

C-Love:
There are few updates here:

1. Adding DWP as a program.
2. Adding a BIG LOOP to make sure that the display limit has not been reached.
3. Adding a BIG LOOP to make sure the benefits months are not issuances that have been archived.
4. Keeping the script from trying to resend cancelled notices. 
5. Adding functionality to put the notice details into a Word Document for every notice created or resent. This happens if 'in person' or 'fax' checkbox is selected.

Updates are in place and have been initially tested. This will be pushed out to all users for testing instead of limited testers.


IlseFerris (approved):
Reviewed as a team. I approve.

Issue #583, opened 2021-09-07
UTILITIES - QI AVS REQUEST Enhancements


MiKaylaH22:
- [ x] Only a message box appears if PRIV. 
    Email will still be sent to QI without data. 
        Send if case is PRIV to QI via email, and identify the case is PRIV. 
	Script end procedure if PRIV.  
- [x ] Only a message box appears if HC is not pending
- [x ] Email will still be sent to QI. 

- [ x] 	Send email to QI and remove msgbox
- [x ] 	Script end procedure if HC not pending
- [x ] If going to PROG you'll want to add in bypass coding if it get stuck on HCRE in edit mode. This happens automatically at times and is inhibiting. 
- [x ] Nothing tied to this code line 108
- [x ] Add this information into the email to QI
- [x ] Enhance mandatory fields: 
-	Add trim on names and SSN (at least 9)
- 	Date handling for DOB
- [x ] Add Closing Issue Documentation when done

PR #607, opened 2021-09-28
#583 UTILITIES - QI AVS REQUEST

MiKaylaH22:
updated to add application date to the email information and ensure that the script is on the right screen in MAXIS.


IlseFerris (changes requested):
Line 301 says if send_email = True then (email sends). Only developer mode is set to false. There should be an if statement or send_email should be defaulted to true.

Issue #187, opened 2020-08-31
NEW SCRIPT - UTILITIES UC Verification Request


IlseFerris:
Hennepin County has UC Liaisons that have access to DEED/Unemployment information for internal verification purposes. A request to support the UC process between the liaisons and the ES staff has been requested. 

Mandora Y. is a SME who will be setting up a meeting to discuss this with our group, and will likely be responsible for the documented process.

IlseFerris:
Faughn has additionally reached out with an outline. She will provide a written process. Not scheduled for now but looking at Q4 2020 or Q1 2021.

IlseFerris:
This is on hold currently until DEED access can be given to more OS staff.

IlseFerris:
Initial process provided by project sponsor.

PR #610, opened 2021-09-29
#187 utilities uc verification request

MiKaylaH22:
Updated for developer mode


IlseFerris (changes requested):
This one is going to have the same issue as #607 pull request. The email will not send if not set to true.

Issue #189, opened 2020-09-01
NEW SCRIPT On Demand case notes for QI


MiKaylaH22:
Taking the template that was created and turning into a script perhaps having it generate emails for corrections along with completion email. also an idea to have it update PROG for missed interview handling and read for possible interview completed case notes. 

Case note script template for notes that are not created by the script
 
Correction needed:
Script - Application Received - not used when pending case
PROG not updated but interview has been completed
Migrant/Seasonal Farmworker field not updated
Script - CAF - used but no interview completed
Script - CAF - not used but approval made
Case not put in to PND2
Worker completed denial for no interview on PND2
Worker completed denial for no interview in ELIG
Interview completed on NOMI day and NOMI was not cancelled
Programs pended on 2 different dates
Clear case note for FS interview requirement not mentioned
Client contact was made no interview offered
Worker completed denial for no interview and did not use PND2

~ Client completed application interview --/--/-- PROG updated ~      
* Completed by previous worker per case note dated: --/--/--          
---                                                                   
EWS Quality Improvement Team                                                         

~ Client completed application interview --/--/-- TYPE & PROG updated ~       
* Completed by previous worker per case note dated: --/--/--
* Per CM 0005.12.03 -aligned CASH application to request for SNAP -           
  Applicants request another assistance program or in some other way amend    
  the application before the agency acts on their initial application.        
  Interview was not completed until 08/28/19 FS still pending                 
*Per CM 0030.03.01 - PROCESSING RCA APPLICATIONS we cannot require another interview for SNAP                                                          
~ Client has not completed CASH F2F application interview ~                       
 * Face to face required because client has not been open on CASH in the last
 12 months  
 * SNAP completed by previous worker per case note dated: 01/30/20             
 ---                                                                           
 EWS Quality Improvement Team                                                  
  
~ Interview not needed for MFIP to SNAP transition  --/--/--  PROG updated ~  
* MFIP to SNAP transition no interview required updated PROG to reflect this                                                          
                                                              
~ Withdrawal completed on PND2 when no interview has been completed ~       
* per case note client requested withdrawal --/--/--
* Client clarified that cash was not requested/needed and the application was 
withdrawn using PACT                                                          
                  
 ~ Application interview ~                                               
* Client is in an IMD FACI and is not required to complete interview  
* Updated PROG to ensure case is processed correctly   

~ Client has not completed application interview ~                            
* Interview is still need for FS, client has 30 days from date of application 
[...]

PR #620, opened 2021-10-04
#189 new script on demand case notes for qi

MiKaylaH22:
removed the email correction and updated the date handling for STAT/PROG


IlseFerris (changes requested):
See email for the deets :)

C-Love:
Option: Client completed application. I selected to NOT have PROG updated through the droplist but the script still put PROG in edit mode though it did not try to take any action. Then the CASE NOTE indicates that PROG was updated. 

Option: Client has not completed application interview. The CASE NOTE has a NOMI space but the dialog didn't require it so it is just blank.

Option: Denied for no interview allows the application date to be less than 30 days.

Issue #475, opened 2021-05-26
ADMIN-ON DEMAND APPLICATION: Remove F2F interview supports


MiKaylaH22:
Remove handling for the face to face interview requirement  and the dropdown in the next action needed

IlseFerris:
No script issues reported by QI users. Closing issue.

PR #621, opened 2021-10-05
#475 & 546 ADMIN-ON DEMAND APPLICATION

MiKaylaH22:


C-Love:
Running notes to review:
- New lines 843-870 can just be deleted - don't need to comment out
- New line 1650 - I think we can remove any verbiage about CASH APP Interview as they should be the same
- New line 1723 - I am not sure what we are looking for here. This was supposed to only indicate MSA cases
- New line 1740 - this shouldn't be an IF statement
- Do we want to add red formatting to other 'NEXT ACTION NEEDED' items - or just wait on the BIG updates?
- No duplication of information in the notes? I am not seeing where it is looking for existing verbiage.

Issue #206, opened 2020-09-25
Build Functions to Read, Update, and Display in Dialog: ADDR, SHEL, and HEST


C-Love:
We have a script called ACTIONS - Shelter Expense Verif Received that will update and note ADDR, SHEL, and HEST when a verification has been received. This is being planed to move into 'Docs Received' at some point. 

Being as there is very specific policy and procedure around updating when an address change is reported but not verified, we should have a script to gather information about a change reported, update panels, and case note. Or maybe we update the current script to work for reports and verifs.

IlseFerris:
SHELTER DEDUCTION CHANGE NC/PC CODES- TE02.08.190

Will still need to determine if we will update ACTIONS - SHELTER EXPENSE VERIFIED or create new script.

IlseFerris:
Reassigned from @MiKaylaH22 to @C-Love since Casey has already done background processing and mapping and has been in communication with project sponsors.

C-Love:
Remember to add functionality for if the case should be transferred.

C-Love:
Handle for Homeowners Insurance information

C-Love:
Also include in the ADDR update to work for inactive cases - which is in issue #252  about updating ADDR for inactive cases. Once the functionality is done we can research where we need it.

C-Love:
Scriptwriter testing is End User Testing

C-Love:
These are ready to be used!!!

PR #639, opened 2021-11-02
#206 MASTER FUNCTIONS LIBRARY - New functions for SHEL, HEST, ADDR reading, display and update

C-Love:
Adding functions to the MASTER Functions Library:

access_HEST_panel - to read and write information from the HEST panel
access_SHEL_panel - to read and write information from the SHEL panel
display_ADDR_information - code for a dialog - to insert into another dialog - to display ADDR information
display_SHEL_information - code for a dialog - to insert into another dialog - to display SHEL information
display_HEST_information - code for a dialog - to insert into another dialog - to display HEST information
display_HOUSING_CHANGE_information - code for a dialog - to insert into another dialog - series of questions and information for any housing change report
navigate_ADDR_buttons - funcctionality to control the dialog from display_ADDR_information function
navigate_SHEL_buttons - funcctionality to control the dialog from display_SHEL_information function
navigate_HEST_buttons - funcctionality to control the dialog from display_HEST_information function
navigate_HOUSING CHANGE_buttons - funcctionality to control the dialog from display_HOUSING_CHANGE_information function
read_total_SHEL_on_case - to read and combine all the SHEL amounts


IlseFerris (changes requested):
HEST updates in both application/renewal & change options are not updating. 
ADDR and SHEL updates occur in the application/renewals option, but not in the change option.

C-Love:
HEST updates hopefully resolved. The issue was in the calling of the function and not so much the function itself. I had an extra boolean. 

Second comment - this is correct - the update functions are only called with the application/renewal part of the script since it is easier to test the updating if the display is a field for field match of the panel. The Change option is to review the dialog display only for when there is a reported change. When built in to an entire script, you would determine what best to do with the reported information - such as CASE:NOTE, Update Panels, or send an Email/Work Item to another area. 

I am going to pull this in with these updates as it appears the FUNC LIB doesn't break when we run this. Additionally I need to not have the Func Lib branched.

Issue #155, opened 2020-08-07
12 Month Contact DAIL Automation


IlseFerris:
Currently OS staff are sending out 12 month contact letters for SNAP recipients who do not have other open programs to inform them of reporting requirements TE02.08.165 - SNAP AGED/DISABLED 12 MONTHS CONTACTS policy. 

Laurie processed only 16 out of 303 message as they had other programs.

IlseFerris:
4 out of 168 messages needed letters for September 2020.

IlseFerris:
13 out of 277 messages needed letter for October 2020.

IlseFerris:
14 out of 167 messages required a 12 month contact letter for November 2020.

IlseFerris:
12 out of 228 messages required a 12 month contact letter for December 2020.

IlseFerris:
14 out of 210 messages sent/required for 12 month contact letter for January 2021.

IlseFerris:
20 out of 204 messages sent/required for 12 month contact letter for February 2021.

IlseFerris:
13 out of 178 messages sent/required for 12 month contact letter for March 2021.

IlseFerris:
5 out of 41 messages sent/required 12 month contact letter for April 2021.

IlseFerris:
0 out of 3 messages sent/required 12 month contact letter for May 2021.

IlseFerris:
1 out of 17 messages sent/required 12 month contact letter for June 2021.

IlseFerris:
7 out of 27 messages sent/required 12 month contact letter for July 2021.

IlseFerris:
26 out of 241 messages sent/required 12 month contact letter for August 2021.

IlseFerris:
19 out of 280 messages sent/required 12 month contact letter for September 2021.

IlseFerris:
15 out of 305 messages sent/required 12 month contact letter for October 2021.

IlseFerris:
Tested for 2 months and working. Closing issue.

PR #640, opened 2021-11-02
#155 dail 12 month contact

IlseFerris:
Initial test ran today. All DAILs were captured. A couple issues with saving the file (resolved) and an issue with not having a MAXIS background check. 

User is Laurie H. Instructions updated locally for her. Script will stay in testing through 12/01/2021 (the next time the run will take place.)


C-Love (approved):
Code looks solid. With limited users I don't see large risks on putting this out.

Issue #646, opened 2021-11-08
DAIL-SCRUBBER NEW HIRE - navigation in DAIL - need to confirm case when first going into stat


MiKaylaH22:
For SDNH matches the navigation is going to the top and navigating into a new case you will know this happens because it will not pull in the member number to the dialog box.

C-Love:
This appears to be fixed.

PR #647, opened 2021-11-08
#646 DAIL - SDNH New Hire has weird navigation things and is sometimes in the wrong case

C-Love:


MiKaylaH22 (approved):
My first review! Reviewed! I was unclear about the AVS but it all looks good.

Issue #642, opened 2021-11-04
UTILITIES - Request PRIV Access - update to advise Teams if on Phone


C-Love:
In speaking to Jen we have decided the best course of action for the UTILITIES - Request PRIV Access script is to stop if the user indicates they are on the phone to alert them that the best way to have an 'urgent' request is to ask for access via Teams to one of the members of Knowledge Now today.

The script will still evaluate the case for being in a basket that does not allow for all users to have access to (foster care or safe at home).

C-Love:
Updates complete.

PR #649, opened 2021-11-09
#642 - UTILITIES - Request Access to PRIV Case

C-Love:
Updates to make the Request access to PRIV case more informative.

Identifies if case is privileged to the foster care team or to a safe at home worker. This changes the email that is sent.

Also recommends using Teams instead of Email for if resident is on the phone.


MiKaylaH22 (approved):
this is great

Issue #645, opened 2021-11-05
NOTES-APPLICATION RECEIVED - remove ability to change the appointment date


MiKaylaH22:
"If interview is being completed please use today's date." This was used for in office same day interview handling and is now messing up notices

C-Love:
Additionally change the e3ditbox to a text for the application date - it cannot be changed.

C-Love:
No issues reported on this change. Update completed.

PR #682, opened 2022-01-02
#645 NOTES - App Recvd - Appt date to be text field with OD explanation

C-Love:
Update made to change the interview date listed in the appointment notice to be a text field instead of an editbox so that no one can change it - ALSO, added a button to link to HSR manual page about On Demand which outlines this whole procedure and the automation.


IlseFerris (approved):
Reviewed code and ran the script. I approve thee!

IlseFerris (approved):
I approve thee!

Issue #708, opened 2022-02-16
Access_ADDR_panel function error - need to clear line of text


C-Love:
Bug reported that the address information is overwritten and partial old data could be maintained.

Update needed to the function to clear text before updating.

C-Love:
Fix is in place - closing issue. The FuncLib did not break.

PR #709, opened 2022-02-16
#708 FuncLib - access_ADDR function - clear line of text

C-Love:
Before writing the information into the panel, the lines need to be cleared. Update to the function to clear the text right away without an if statement.


IlseFerris (approved):
I approve three!

Issue #728, opened 2022-03-02
BUG on UTILITIES - Update Check Dates where the same paycheck keeps getting added


C-Love:
This is specific to semi-monthly pay sources. These calculations are the most complicated in this script when moving through the list of pay dates. 

It appears the issue is that the script at one point is comparing a number to a string for the day of pay within a month - particularly if that was determined by the dialog droplist box selection which was initially unknown.

C-Love:
This is done - no new issues reported

PR #731, opened 2022-03-02
#728 - UTILITIES - Update Check Dates - bug on determining semi-monthly pay dates

C-Love:
Review how the script assesses the dates to move through the sequence of dates for income sources that are paid semi-monthly. 

A large part of this was also ensuring the script is recognizing that the days indicated as the regular monthly pay dates are viewed as numbers and not strings for correct comparison.


IlseFerris (approved):
Tested on training case 305411. Updated CM an CM + 1 as intended for semi-monthly. I approve thee!

Issue #711, opened 2022-02-17
ACTIONS - Earned Income Budgeting - combine paychecks from the same day


C-Love:
When multiple checks from the same day and same employer are received the script does not correctly calculate or update in the panel. 

Enhancement plan:
 - Encourage workers to enter checks separately in the EIB dialog.
 - Combine check amounts from the same day when calculating and entering information into MAXIS JOBS panels
 - Keep checks separated in the CASE NOTE
 - Add a single line to CASE NOTE of the combination in JOBS if that happens. 

This needs additional discovery on HOW to get this done.

C-Love:
@IlseFerris  - I realized after our discussion that I would like to add some verbiage on the paycheck entry dialog to combine checks when entering for now to prevent errors. Do you think that is a value added stop gap measurement for this issue?

IlseFerris:
@C-Love go for launch on this.

C-Love:
No errors reported. Closing.

PR #732, opened 2022-03-03
#711 ACTIONS - Earned Income Budgeting - verbiage added to dialog for multiple checks

C-Love:
The script does not behave correctly if multiple checks from one date are entered separately. We will eventually be updating this script to handle for that correctly, but for the time being, we are adding some words to the check entry dialog to let people know to combine these checks.


IlseFerris (approved):
I approve thee!

Issue #725, opened 2022-03-01
function convert_digit_to_excel_column causing inhibiting errors for Review Report


IlseFerris:
The bounds of the function have been reached with ADMIN - REVIEW REPORT for 02/22 renewals. Extended to support excel columns that are greater than 104 to resolve issue. Will discuss with team to decide longevity updates.

IlseFerris:
Update - adding supports into the H column. @IlseFerris will update. @C-Love to review.

IlseFerris:
Function is working, no issues reported by LH, primary user. Closing issue.

PR #750, opened 2022-03-07
#725 function convert_digit_to_excel_column: added supports through H…

IlseFerris:
… column


C-Love (approved):
Numbers look good. I think this is good to go.

PR #751, opened 2022-03-07
ACTIONS - grh rate 2, If statement never true

BMarshyyy:
Counties using the DHS scripts require their global variables county code to have a lower case x 
(example: x105 used and not X105)

the var "current_county" reads from maxis as "X105". 
Using replace on line 123 resolves the issue of a lower case x in the global variables without needing to change other DHS scripts to make this script work. This should also have no effect on Hennepin running the code.

### DIFF
--- a/actions/add-grh-rate-2-to-mmis.vbs
+++ b/actions/add-grh-rate-2-to-mmis.vbs
@@ -120,9 +120,8 @@ If grh_status <> "ACTV" then
 	script_end_procedure("GRH case status is " & grh_status & ". The script will now end.")
 End if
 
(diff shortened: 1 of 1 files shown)


IlseFerris (changes requested):
This is a great enhancement idea, especially for our out-of-county folks. Generally we would use UCASE(worker_county_code) vs. the replace method. This follows our style of scripting in other projects. If you would remove Line 123 and update line 125 I'll approve this pull request.

BMarshyyy (commented):
Line 123 has been removed and line 124 has "worker_county_code" replaced with "UCase(worker_county_code)".

IlseFerris (approved):

Issue #700, opened 2022-01-27
DEU - BULK MATCH CLEARED - bring the script out of retirement 


MiKaylaH22:
This was initially established as a QTR 1 pivot  due to changes in circumstance it is no longer identified as a PIVOT but a long term assistant to the HSS's process/procedure. This script will reduce the amount of copy and paste that is manually done by the HSS of the team. We will limit the use to only the HSS.

IlseFerris:
Part 1 of issue - release original version for DEU. This will come out today. 
Part 2 of issue - update script with planned enhancements (@MiKaylaH22 will list these in a comment), which includes restricting the usage to MH only.

MiKaylaH22:
No bugs reported

PR #757, opened 2022-03-11
#700 DEU BULK MATCH CLEARED -reinstate

MiKaylaH22:
a pivot created a need to reinstate the BULK MATCH CLEARED, with limited access for HSS only


IlseFerris (changes requested):
@MiKaylaH22 I have some changes requested and questions labeled with ??? in the code. Also I am not able to test this so that's a must too.

Issue #749, opened 2022-03-07
BUG on NOTES - CAF - Verifs Needed NOTE is not being created


C-Love:
This has been reported intermittently that when a Verif CNOTE does not yet exist and verifications are entered, the note is not created. 

This does not happen all the time and the SYW file indicates the logic in the script is correct. 

I believe this is likely to do with MAXIS navigation. 

I have a couple ideas:
1, More Discovery - add a report out to SYW file if the NOTE was attempted - this would be in each of the optional CASE NOTEs
2. Add a 'Back to Self' between NOTEs attempts - right after the 'PF3' for each NOTE
3. Add a loop to each optional CASE NOTE that reads the top line for the right header and tries again until found.

These second 2 would add some error proofing and may resolve the issue by adding stability but we wouldn't know until the reports don't come in about it.

C-Love:
Try option 1 and 2 together - creating more awareness and a little bit of code stability.

C-Love:
Work completed on this issue. No new errors reported. Hopefully we don't hear of this again.

PR #763, opened 2022-03-16
#749 - NOTES - CAF - Stability added for CASE NOTE functionality

C-Love:
Added:
1. Back to self between each note to ensure MAXIS is moving in the right way.
2. report out about the attempts of the script in creating CASE NOTE in the save your work and script run lowdown.
3. Found a string/boolean issue with restore your work that might actually be the cause - reviewed variables and added resolution code for the boolean variables. 
4. Removed MFIP DVD functionality
5. Added a script run lowdown detail to dindicate if a script run is using restore your work for future error report review

All functionality is ready for review.

### DIFF
--- a/notes/caf.vbs
+++ b/notes/caf.vbs
@@ -9310,6 +9310,7 @@ If HC_checkbox = checked Then
     Call write_variable_in_CASE_NOTE(worker_signature)
 
     PF3
+    Call back_to_SELF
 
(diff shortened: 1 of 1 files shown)


IlseFerris (approved):
Didn't run, but code looks solid. I also see that you tested this like a dozen times. I approve thee!

Issue #771, opened 2022-03-24
NOTES - CAF Bug when words are added to the app month asset field


C-Love:
The Expedited Determination functionality cannot handle if the information entered into the field about Liquid Assets on Dialog 7 is not numeric - the script things break. 

I need logic to ignore this information if it is not numeric.

C-Love:
This is done

PR #777, opened 2022-03-29
#771 NOTES - CAF - ensure all exp det variables are numbers

C-Love:
The is the resolution to the bug that CK reported last week. This has only been reported once and is because non-numeric information was entered in a field that was expected to be only numeric. 

This functionality will 'zero out' any information gathered from fields that are not numeric.


IlseFerris (approved):
Logic looks sound to me. I approve thee!

Issue #697, opened 2022-01-20
Replace APPPLYMN applications with ONLINE applications in script


MiKaylaH22:
ApplyMN will no longer be available after January 31, 2022. New applications are no longer accepted on ApplyMN. 

Now through January 31, ApplyMN will only be available for people with partial or incomplete applications. On February 1, any partial or incomplete ApplyMN applications will no longer be available. If you go to the ApplyMN page (on or after this date) you will be redirected to MNbenefits. 

Because we have a delay in APPLs being processed this will not sundown until 02/14/22 

- [x] utilities/lost-applymn.vbs - this one needs to be able to look back retroactive
- [x] misc/complete-task.vbs
- [x] misc/tasks-dash.vbs
- [x] notices/eligibility-notifier.vbs
- [x] notes/application-received.vbs
- [x] notes/application-check.vbs
- [x] retired-scripts/waived-er-interview-screening.vbs
- [x] actions/full-interview.vbs

MiKaylaH22:
Ilse may take this one

IlseFerris:
No issues or updates reported/requested. Closing issue.

Issue #628, opened 2021-10-12
Update UTILITIES - LOST APPLYMN: Name of script and functionality


MiKaylaH22:
Replace name and update documentation to ensure to support MNbenefits vs, ApplyMN applications, and additional support functionality/updates.  Name of the script will now be APPLICATION INQUIRY.

Functionality
- [x] Add CASE ASSIGMENT navigation button on SPO. 
- [x] Auto-filling date of online application will need to be updated (pulling in date inconsistently)
- [x] Replacement of ApplyMN to MNbenefits

For the Name Change
- [x]  locally-installed-files/redirect-utilities-request-lost-applymn.vbs - name change
- [x] utilities - lost applymn - name change
- [x] Update power pads (general and specialty)
- [x] /COMPLETE LIST OF SCRIPTS.vbs
- [x] Power Pad button name update in SCRIPTS FOR SUPPORT section
- [x] Changelog announcement in FuncLib

C-Love:
Need process steps in writing and in the HSR manual.

IlseFerris:
Confirmed process for MNBenefits and ApplyMN is the same. Putting this in On Deck status. This will be a great help to CA team. 
Will also need to update the power pad button name in the 'scripts for support' section.

IlseFerris:
Name will be updated to APPLICATION INQUIRY

IlseFerris:
Adding button to CASE ASSIGNMENT SharePoint Page.

IlseFerris:
Aside from the unintended updating of the power pad to remove tags & favorites (which has been resolved, thank you @C-Love, no issues have been reported. Closing issue.

PR #778, opened 2022-03-29
#697 Removed APPLYMN verbiage from scripts

IlseFerris:
UTILITIES - LOST APPLYMN name will change with the work in #628

### DIFF
--- a/utilities/lost-applymn.vbs
+++ b/utilities/lost-applymn.vbs
@@ -68,8 +68,11 @@ function search_for_info_in_note(date_of_app, applymn_confirmation_number, name_
 
 			case_note_header = trim(case_note_header)							'reformat the case note
 			If case_note_date <> "        " Then case_note_date = DateAdd("d", 0, case_note_date)		'make the date a date
-
-			If case_note_header = "~ Application Received (ApplyMN)" or case_note_header = "~ Application Received (MNbenefi" Then		'If it finds an application received case note for an online application
+            'msgbox case_note_header
+            'If it finds an application received case note for an online application
+			If case_note_header = "~ Application Received (ApplyMN)" or _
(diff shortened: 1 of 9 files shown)


utilities/lost-applymn.vbs
```diff
@@ -68,7 +69,7 @@ function search_for_info_in_note(date_of_app, applymn_confirmation_number, name_
 			case_note_header = trim(case_note_header)							'reformat the case note
 			If case_note_date <> "        " Then case_note_date = DateAdd("d", 0, case_note_date)		'make the date a date
 
-			If case_note_header ="~ Application Received (ApplyMN)" Then		'If it finds an application received case note for an ApplyMN
+			If case_note_header = "~ Application Received (ApplyMN)" or case_note_header = "~ Application Received (MNbenefi" Then		'If it finds an application received case note for an online application
```
C-Love, line 72: You have updated Application Received to use 'Mnbenefits' but this will look for 'MNbenefits' - it won't be able to find a match. This should probably have options to have a space and not and have it coded to a UCase so that any option works.

C-Love (changes requested):
I generally find MNBenefits or MNbenefits more readable than Mnbenefits. I am not willing to die on this hill though so I don't really care. I do wish someone else would decide this convention - within Hennepin - then we can follow it.

There is a case mismatch in two scripts in which one reads from the other.

utilities/lost-applymn.vbs
IlseFerris: Updates made. I also reached out to Faughn to see get HC information about the naming convention.

MiKaylaH22 (approved):
It looks good, I do agree with Casey about the naming convention MNBenefit makes the most sense

Issue #779, opened 2022-03-29
Update naming convention for MNbenefits


IlseFerris:
Naming convention has been confirmed as MNbenefits. The following scripts will need to be updated with the new naming convention: 

- [x] MASTER FUNCTIONS LIBRARY
- [x] ADMIN - IND APPT NOTICE
- [x] ADMIN - IND NOMI
- [x] ADMIN - IND RECERT NOTICES
- [x] ADMIN - ON DEMAND WAIVER APPLICATIONS
- [x] ADMIN - ON DEMAND NOTES
- [x] ADMIN - REVIEW REPORT
- [x] ADMIN - TRACK AUTOCLOSE PAYMENTS 
- [x] NOTES - APPLICATION RECEIVED 
- [x] NOTES - CAF
- [x] NOTES - EXPEDITED DETERMINATION
- [x] NOTES - INTERVIEW
- [x] NOTICES - ADD WCOM

There are a couple of scripts in the desert island scripts that would also need to be updated, but those will be updated on 4/1 as part of the monthly maintenance of our script project.

IlseFerris:
No issues reported. Closing issue.

PR #785, opened 2022-03-31
#779 MNbenefits naming convention

IlseFerris:
All scripts and functions updated. Ready for review.


C-Love (approved):
Looks good - doesn't look like any functionality impact.

MiKaylaH22 (changes requested):
Caf line 795 this is spelled theis 
Notes Exp 771 this is spelled theis
Interview 745 this is spelled theis
also there is a missing space between document.Advise everything else looks great but since we are doing things right

Issue #801, opened 2022-04-11
ADMIN - TASK BASED DAIL CAPTURE: bug when case loads don't have DAIL info


IlseFerris:
No handling for cases when the DAIL messages selected don't exist in the DAIL/PICK. This is in the DAIL Decimator, but was removed when the update to this process was made (that was me, I updated that...sigh....).

IlseFerris:
No issues reported by user. Closing issue.

PR #802, opened 2022-04-12
#801 ADMIN - TASK-BASED DAIL CAPTURE: Added additional handling

IlseFerris:
for cases that have no DAIL messages in the DAIL/PICK that meet the criteria of DAIL types the script is looking for.

Quick fix, that appears to be working. For some reason with the DAIL/PICK vs DAIL/DAIL the # of DAILs isn't always exiting the do. DAIL/PICK is not the typical way we look for these messages, but seems more stable, just different than DAIL/DAIL.


IlseFerris:
OK @C-Love and @MiKaylaH22 I added closing project documentation so this one is ACTUALLY ready now :)

C-Love (approved):
This looks solid

Issue #769, opened 2022-03-24
ACTIONS - TRANSFER CASE - need discovery on handling the secondary worker


MiKaylaH22:
Information: ERROR MESSAGE DISPLAYED DUE TO 2ND SERVICE WORKER ID 
Closing message: Case did not transfer, please review for where the error occured.

MiKaylaH22:
Discovery completed and decided that removal of secondary worker is best choice for now

MiKaylaH22:
Discussed at meeting and because we clear this line of text, until more documentation is received around this issue.

Issue #759, opened 2022-03-14
ACTIONS - TRANSFER CASE: review handling for primary worker and secondary worker


MiKaylaH22:
case is not transferring when 2nd worker is listed case note is still writing despite the transfer not being successful

IlseFerris:
Closed as a won't fix. Secondary workers are deleted during the transfer.

Issue #466, opened 2021-05-17
ACTIONS - Transfer Case - Restrict transfer of cases to CCL with the script


C-Love:
There is a batch process to correctly transfer cases to CCL at the appropriate time. Workers should NOT be transferring cases on an individual case basis. 

https://hennepin.sharepoint.com/teams/hs-es-manual/SitePages/Closing_Case_Records.aspx

This is manual reference to support this.

Update the script to prevent transfer to x127CCL with messaging to the worker that these transfers should not be happening.

MiKaylaH22:
Closing as #393 remains open for testing

Issue #443, opened 2021-04-21
ACTIONS - TRANSFER CASE: audit out-of-county option for case note information


IlseFerris:
Per POLI TEMP TE02.08.095 CASE NOTE III: CLAIMS/SYSTEMS/TRANSFERS the following information should be included in the case note when transferring to another county:                                 
- Note date of client's move
- New address
- Whether the client has entered or left excluded time.
- List all pending verifications and case actions.     
- Note any unresolved DAIL messages or IEVS matches.              
- Note any expected changes in household's circumstances.
-  Indicate if any income reports/recertification are outstanding.             
-   spelling on the word occurred 
- link to sharepoint page

MiKaylaH22:
Closing as #393 remains open for testing

PR #809, opened 2022-04-15
ACTIONS TRANSFER CASE #393, 442, 443, 466, 759 & 769 multi issue branch 

MiKaylaH22:
- [x] #769 - Information: ERROR MESSAGE DISPLAYED DUE TO 2ND SERVICE WORKER ID this issue is obsolete and can be closed 
- [x] #759 - case is not transferring when 2nd worker is listed case note is still writing despite the transfer not being successful
  - removed secondary worker 
- [x] #466 - Restrict transfer of cases to CCL with the script 
  - script end with message explaining in 4 months this will happen via automation
- [x]  #442 Move the NOTE to after the XFER confirmed 
  - Move the CASE:NOTE to a nested IF statement that indicates the XFER was successful but once we xfer we cannot case note anymore 
- [x] #443 audit out-of-county option for case note information - no internal case note needed
- [x] #393 Should this case note in all situations? -nope

### DIFF
--- a/actions/transfer-case.vbs
+++ b/actions/transfer-case.vbs
@@ -65,7 +65,7 @@ back_to_self                                        'added to ensure we have the
 
 EMReadScreen worker_number, 7, 22, 8                'reading the current workers number '
 EMWriteScreen MAXIS_case_number, 18, 43             'writing in the case number so that if cancelled, the worker doesn't lose the case number.
-msgbox worker_number
+
 closing_message = "Transfer case is complete."      'setting up closing_message variable for possible additions later based on conditions
 
 BeginDialog Dialog1, 0, 0, 201, 85, "Match Cleared"
@@ -116,12 +116,11 @@ call navigate_to_MAXIS_screen("REPT", "USER")
 EMWriteScreen transfer_to_worker, 21, 12
 TRANSMIT
 DO
-    EMReadScreen worker_found, 15, 24, 2
-    msgbox worker_found & " " &  inactive_worker
+    EMReadScreen error_message, 75, 24, 2
     EMReadScreen inactive_worker, 8, 7, 38
     IF inactive_worker = "INACTIVE" THEN MsgBox "The worker or agency selected is not active. Please try again."
-    IF worker_found = "NO WORKER FOUND" THEN MsgBox "The worker or agency selected does not exist. Please try again."
-LOOP UNTIL inactive_worker <> "INACTIVE" AND worker_found <> "NO WORKER FOUND"
+    msgbox error_message & " " &  inactive_worker
+LOOP UNTIL inactive_worker <> "INACTIVE"
 transfer_out_of_county = FALSE                                          'setting varible to false'
 IF left(transfer_to_worker, 4) <> "X127" THEN transfer_out_of_county = TRUE 'setting the out of county BOOLEAN'
 
@@ -135,18 +134,18 @@ IF worker_agency_name = "" THEN 						'If we are unable to find the alias for th
 	name_length = len(worker_agency_name)
 	comma_location = InStr(worker_agency_name, ",")
 	worker_agency_name = right(worker_agency_name, (name_length - comma_location)) & " " & left(worker_agency_name, (comma_location - 1)) 'this section will reorder the name of the worker since it is stored here as last, first. the comma_location - 1 removes the comma from the "last,"
-END IF
-EMReadScreen mail_addr_line_one, 43, 9, 27
-	mail_addr_line_one = trim(mail_addr_line_one)
-EMReadScreen mail_addr_line_two, 43, 10, 27
-	mail_addr_line_two = trim(mail_addr_line_two)
-EMReadScreen mail_addr_line_three, 43, 11, 27
-	mail_addr_line_three = trim(mail_addr_line_three)
-EMReadScreen mail_addr_line_four, 43, 12, 27
-	mail_addr_line_four = trim(mail_addr_line_four)
-EMReadScreen worker_agency_phone, 14, 13, 27
-EMReadScreen worker_county_code, 2, 15, 32
 
+    EMReadScreen mail_addr_line_one, 43, 9, 27
+    	mail_addr_line_one = trim(mail_addr_line_one)
+    EMReadScreen mail_addr_line_two, 43, 10, 27
+    	mail_addr_line_two = trim(mail_addr_line_two)
+    EMReadScreen mail_addr_line_three, 43, 11, 27
+    	mail_addr_line_three = trim(mail_addr_line_three)
+    EMReadScreen mail_addr_line_four, 43, 12, 27
+    	mail_addr_line_four = trim(mail_addr_line_four)
+    EMReadScreen worker_agency_phone, 14, 13, 27
+    EMReadScreen worker_county_code, 2, 15, 32
+END IF
 transfer_message = ""            'some defaults
 transfer_case = False
 action_completed = True
@@ -216,8 +215,8 @@ ELSE
      ButtonGroup ButtonPressed
      OkButton 235, 260, 50, 15
      CancelButton 290, 260, 50, 15
-     PushButton 135, 5, 65, 15, "SPEC/XFER", nav_SPEC_XFER_button
-     PushButton 205, 5, 65, 15, "POLI/ TEMP", nav_POLI_TEMP_button
+     PushButton 135, 5, 65, 15, "SPEC/XFER", XFER_button
+     PushButton 205, 5, 65, 15, "POLI/TEMP", POLI_TEMP_button
      PushButton 275, 5, 65, 15, "USEFORM", useform_xfer_button
      Text 5, 10, 60, 10, "Client Move Date"
      Text 5, 30, 55, 10, "Excluded time?"
@@ -243,6 +242,14 @@ ELSE
 			err_msg = ""
 			Dialog Dialog1
 			cancel_confirmation
+            If ButtonPressed = useform_xfer_button Then               'Pulling up the hsr page if the button was pressed.
+                run "C:\Program Files (x86)\Microsoft\Edge\Application\msedge.exe https://aem.hennepin.us/DocumentManager/docm1649100617025/dad342433412ab721a67d46f95a3d1c1?type=YXBwbGljYXRpb24vcGRm"
+                err_msg = "LOOP"
+            Else                                                'If the instructions button was NOT pressed, we want to display the error message if it exists.
+    		    IF err_msg <> "" THEN MsgBox "*** NOTICE!***" & vbNewLine & err_msg & vbNewLine
+            End If
+	        If ButtonPressed = POLI_TEMP_button then call navigate_to_MAXIS_screen("POLI", "TEMP")
+            Call MAXIS_dialog_navigation()
 	      	IF transfer_reason = "" THEN err_msg = err_msg & vbNewLine & "Please enter a reason for transfer."
             IF excluded_time_dropdown = "Yes" AND isdate(excluded_date) = False THEN MsgBox "Please enter a valid date for the start of excluded time or double check that the client's absense is due to excluded time."
 			IF isdate(move_date) = False THEN MsgBox "Please enter a valid date for client move."
@@ -447,7 +454,7 @@ ELSE
 END IF ' the big one'
 
 IF end_msg <> "" Then
-    If transfer_case = True Then closing_message = closing_message & vbCr & vbCr & "Case transfer has been completed to x127" & transfer_to_worker
+    If transfer_case = True Then closing_message = closing_message & vbCr & vbCr & "Case transfer has been completed to: " & transfer_to_worker
 ELSE
     closing_message = closing_message & vbCr & vbCr & "FAILED CASE TRANSFER:" & vbCr & end_msg
 END IF
--- a/actions/transfer-case.vbs
+++ b/actions/transfer-case.vbs
@@ -130,7 +130,12 @@ DO
     worker_number = Ucase(worker_number)
 
     '----------Checks that the worker or agency is valid---------- 'must find user information before transferring to account for privileged cases.
-    'transfer_to_worker = ucase(transfer_to_worker)
+
+    CALL navigate_to_MAXIS_screen_review_PRIV("CASE", "CURR", is_this_priv) ' need discovery on priv cases for xfer handling'
+    IF is_this_priv = TRUE THEN script_end_procedure("This case is privileged, the script will now end.")
+
+    IF transfer_to_worker = "X127CCL" THEN script_end_procedure("This case is will be transferred via an automated script after being closed for 4 months, the script will now end.")
+
     call navigate_to_MAXIS_screen("REPT", "USER")
     EMWriteScreen transfer_to_worker, 21, 12
     TRANSMIT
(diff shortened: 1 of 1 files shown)


actions/transfer-case.vbs
```diff
@@ -51,84 +51,124 @@ changelog = array()
...
-CALL changelog_update("11/09/2020", "No issues with SPEC/MEMO for out-of-county cases. SIR Announcement from 11/05/20 stated an issue was identified. Hennepin County's script project is seperate from DHS's script project. We are not experiencing the reported issue. Thank you!", "Ilse Ferris, Hennepin County")
-CALL changelog_update("10/20/2020", "Updated link for out-of-county use form.", "Ilse Ferris, Hennepin County")
-CALL changelog_update("10/16/2020", "Added closing message box with information if a case is identified as being transferred to the MNPrairie Bank of counties.", "Ilse Ferris, Hennepin County")
-CALL changelog_update("04/20/2020", "Rewrite.", "MiKayla Handley, Hennepin County")
-CALL changelog_update("03/19/2019", "Added an error reporting option at the end of the script run.", "Casey Love, Hennepin County")
-CALL changelog_update("12/29/2017", "Coordinates for sending MEMO's has changed in SPEC/MEMO. Updated script to support change.", "Ilse Ferris, Hennepin County")
-call changelog_update("11/28/2016", "Initial version.", "Charles Potter, DHS")
+call changelog_update("03/28/2022", "Initial version.", "MiKayla Handley, Hennepin County.")
```
C-Love, line 54: We don't want to lose the whole changelog - sorry, I wasn't clear on this with the review we did. Starting it over can mess up the changelog file that are on worker's documents folder. You just want to add the changes to a new line in the changelog.

actions/transfer-case.vbs
```diff
@@ -51,84 +51,124 @@ changelog = array()
...
+    LOOP UNTIL are_we_passworded_out = false					'loops until user passwords back in
+
+    transfer_to_worker = trim(transfer_to_worker)                'formatting the information entered in the dialog
+    transfer_to_worker = Ucase(transfer_to_worker)               'making sure we are capital in all things'
+    worker_number = trim(worker_number)
+    worker_number = Ucase(worker_number)
+
+    '----------Checks that the worker or agency is valid---------- 'must find user information before transferring to account for privileged cases.
+
+    CALL navigate_to_MAXIS_screen_review_PRIV("CASE", "CURR", is_this_priv) ' need discovery on priv cases for xfer handling'
```
C-Love, line 134: Recommend moving this (line 134-137) outside of this larger loop so that the script doesn't bounce around in MAXIS too much if we have to loop.

actions/transfer-case.vbs
```diff
@@ -138,466 +178,296 @@ EMReadScreen mail_addr_line_four, 43, 12, 27
...
-	IF forms_to_swkr = "Y" THEN call write_variable_in_case_note("* Copy of SPEC/MEMO sent to social worker.")
+    Call write_bullet_and_variable_in_CASE_NOTE("Actions to be taken", action_to_be_taken)
+    Call write_bullet_and_variable_in_CASE_NOTE("Requested verifications", requested_verifs)
+    Call write_bullet_and_variable_in_CASE_NOTE("Expected changes in household's circumstances:", expected_changes)
+    Call write_bullet_and_variable_in_CASE_NOTE("Other notes", other_notes)
+    IF mets_status_dropdown = "Active" THEN Call write_bullet_and_variable_in_CASE_NOTE("* Client is active on HC through METS case number:", METS_case_number)
+    call write_bullet_and_variable_in_case_note("Client move date", client_move_date)
+    call write_bullet_and_variable_in_case_note("Change report sent", date) 'defaulting information '
+    call write_bullet_and_variable_in_case_note("Case file sent:", date) 'defaulting information '
+    IF excluded_time = "Yes" THEN
```
C-Love, line 332: There is still some variable management needed here. We had talked about changing it to:
```
IF excluded_time_dropdown = "Yes" THEN
call write_bullet_and_variable_in_case_note("Excluded Time" , "Yes, Begins " & excluded_date)
ELSEIF excluded_time_dropdown = "No" THEN
call write_bullet_and_variable_in_case_note("Excluded Time", excluded_time_dropdown)
END IF
```

actions/transfer-case.vbs
```diff
@@ -138,466 +178,296 @@ EMReadScreen mail_addr_line_four, 43, 12, 27
...
+        excluded_time = excluded_time & ", Begins " & excluded_date
+        call write_bullet_and_variable_in_case_note("Excluded Time" , excluded_time)
+        excluded_time = "Yes"
+    ELSEIF excluded_time = "No" THEN
+        call write_bullet_and_variable_in_case_note("Excluded Time", excluded_time)
+    ELSEIF excluded_time = "" THEN
+        call write_variable_in_case_note("* Excluded Time: N/A")
+    END IF
+    IF ma_status = "ACTIVE" THEN
```
C-Love, line 341: hc CFR and Cash CFR variables are not consistent. The dialog variables and XFER functionality match, but it is different functionality in the CASE/NOTE.

actions/transfer-case.vbs
```diff
@@ -138,466 +178,296 @@ EMReadScreen mail_addr_line_four, 43, 12, 27
...
-			IF (hc_status_check = "ACTV" AND excluded_time = "") THEN MsgBox "Please select whether the client is on excluded time."
-			IF manual_cfr_cash_checkbox = CHECKED AND cash_cfr_no_change_checkbox = CHECKED THEN MsgBox ("Please select whether the CFR for CASH is changing or not. Review input.")
-			IF manual_cfr_hc_checkbox = CHECKED AND hc_cfr_no_change_checkbox = CHECKED THEN MsgBOx ("Please select whether the CFR for HC is changing or not. Review input.")
-			If mets_active_checkbox = CHECKED and METS_case_number = "" then err_msg = err_msg & vbNewLine & "* Please enter a METS case number."
-			If err_msg <> "" THEN MsgBox "*** NOTICE!!! ***" & vbNewLine & err_msg & vbNewLine		'error message including instruction on what needs to be fixed from each mandatory field if incorrect
-		LOOP UNTIL err_msg = ""									'loops until all errors are resolved
+            IF ButtonPressed = POLI_TEMP_button THEN CALL view_poli_temp("TE02", "08", "095", "") 'TE02.08.095' there is no forth variable
+            IF ButtonPressed = XFER_button Then CALL MAXIS_dialog_navigation()
```
C-Love, line 283: The If statement isn't needed here, unless you are having some dialog display issues.

actions/transfer-case.vbs
```diff
@@ -138,466 +178,296 @@ EMReadScreen mail_addr_line_four, 43, 12, 27
...
+            IF excluded_time_dropdown = "Yes" AND isdate(excluded_date) = False THEN err_msg = err_msg & vbNewLine & "* Please enter a valid date for the start of excluded time or double check that the client's absense is due to excluded time."
+			IF isdate(client_move_date) = False THEN  err_msg = err_msg & vbNewLine & "* Please enter a valid date for client move."
+			IF (ma_status = "ACTIVE" AND excluded_time_dropdown = "No") THEN  err_msg = err_msg & vbNewLine & "* Please select whether the client is on excluded time."
+			IF manual_cfr_cash_checkbox = CHECKED AND cash_cfr_no_change_checkbox = CHECKED THEN err_msg = err_msg & vbNewLine & "* Please select whether the CFR for CASH is changing or not. Review input."
+			IF manual_cfr_hc_checkbox = CHECKED AND hc_cfr_no_change_checkbox = CHECKED THEN  err_msg = err_msg & vbNewLine & "* Please select whether the CFR for HC is changing or not. Review input."
+			IF (mets_status_dropdown = "Active" and METS_case_number = "") then err_msg = err_msg & vbNewLine & "* Please enter a METS case number."
+			IF err_msg <> "" THEN MsgBox "*** NOTICE!***" & vbNewLine & err_msg & vbNewLine
+		Loop until err_msg = ""
```
C-Love, line 293: I think you might end up leaving the dialog loop when pressing one of the policy buttons if there are no errors found in the variable review. There needs to be something to ensure that the 'Loop until  err_msg = ""' doesn't exit if the button pressed isn't the 'OK Button'

actions/transfer-case.vbs
```diff
@@ -138,466 +178,296 @@ EMReadScreen mail_addr_line_four, 43, 12, 27
...
+    Call write_variable_in_SPEC_MEMO(worker_agency_name)
+    Call write_variable_in_SPEC_MEMO(mail_addr_line_one)
+    Call write_variable_in_SPEC_MEMO(mail_addr_line_two)
+    Call write_variable_in_SPEC_MEMO(mail_addr_line_three)
+    Call write_variable_in_SPEC_MEMO(mail_addr_line_four)
+    Call write_variable_in_SPEC_MEMO(worker_agency_phone)
+    Call write_variable_in_SPEC_MEMO("Domestic violence brochures are available at https://edocs.dhs.state.mn.us/lfserver/Public/DHS-3477-ENG.")
+    Call write_variable_in_SPEC_MEMO("You can also request a paper copy.  Auth: 7CFR 273.2(e)(3).")
+
+    IF forms_to_arep = "Y" THEN EMWriteScreen "x", arep_row, 10     'If forms_to_arep was "Y" (see above) it puts an X on the row ALTREP was found.
```
C-Love, line 313: These two lines are not needed. The function 'start_a_new_spec_memo' will select these if they are 'Y' when starting the MEMO.

actions/transfer-case.vbs
```diff
@@ -138,466 +178,296 @@ EMReadScreen mail_addr_line_four, 43, 12, 27
...
+    IF forms_to_swkr = "Y" THEN EMWriteScreen "x", swkr_row, 10     'If forms_to_arep was "Y" (see above) it puts an X on the row ALTREP was found.
+    TRANSMIT                                       'TRANSMITs to start the memo writing process
```
C-Love, line 315: This transmit shouldn't be needed here. I'm not sure what this would do.

actions/transfer-case.vbs
```diff
@@ -138,466 +178,296 @@ EMReadScreen mail_addr_line_four, 43, 12, 27
...
+    PF4'save and exit
+
+    '----------The case note of the reason for the XFER----------
     start_a_blank_CASE_NOTE
-    Call write_variable_in_CASE_NOTE("~ Case transferred to " & servicing_worker & " ~")
-    Call write_bullet_and_variable_in_CASE_NOTE("Active programs", active_programs)
-    Call write_bullet_and_variable_in_CASE_NOTE("Pending programs", pend_programs)
+    Call write_variable_in_CASE_NOTE("~ Case transferred to " & transfer_to_worker & " ~")
+    Call write_bullet_and_variable_in_CASE_NOTE("Active programs", list_active_programs)
+    Call write_bullet_and_variable_in_CASE_NOTE("Pending programs", list_pend_programs)
```
C-Love, line 322: the variable 'list_pend_programs' does not match the variable that is called in the 'determine_program_and_case_status_from_case_curr' function. This will always be blank.

actions/transfer-case.vbs
```diff
@@ -138,466 +178,296 @@ EMReadScreen mail_addr_line_four, 43, 12, 27
...
+            EMReadScreen new_primary_worker, 7, 21, 16
+            IF new_primary_worker <> transfer_to_worker Then           'if it is not the transfer_to_worker - the transfer failed.
+                transfer_case = False
+                end_msg = "Transfer of this case to " & transfer_to_worker & " has failed."
+            End If
+        END IF
+    END IF ' confirming the xfer worked and we left the panel'
+END IF ' the big one'
+
+IF end_msg <> "" Then
```
C-Love, line 426: These if statements are a little redundant. If transfer case is false, there will always be information in 'end_msg' because they are defined in this way right next to each other. You might not need the 'transfer_case' boolean, or the if statement might be able to be just a little cleaner.

actions/transfer-case.vbs
```diff
@@ -51,84 +51,124 @@ changelog = array()
...
-	EMReadScreen inactive_worker, 8, 7, 38
-	IF inactive_worker = "INACTIVE" THEN MsgBox "The worker or agency selected is not active. Please try again."
-	IF worker_found = "NO WORKER FOUND" THEN MsgBox "The worker or agency selected does not exist. Please try again."
-LOOP UNTIL inactive_worker <> "INACTIVE" AND worker_found <> "NO WORKER FOUND"
+    EMReadScreen error_message, 75, 24, 2
+    error_message = trim(error_message)
+    EMReadScreen inactive_worker, 8, 7, 38
+    IF inactive_worker = "INACTIVE" THEN
+        active_worker_found = false
+        MsgBox "Please review the worker and try again. " & error_message
```
C-Love, line 148: There is no guarantee that the 'error_message' has any relevant information here. The message here should be specific to the inactive worker status.

actions/transfer-case.vbs
```diff
@@ -51,84 +51,124 @@ changelog = array()
...
+    END IF
+    IF error_message = "NO WORKER FOUND WITH THIS ID" THEN
+        active_worker_found = false
+        MsgBox "Please review the worker and try again. " & error_message
```
C-Love, line 152: The format of this message is clunky, the details would all be on one line and it isn't clear that the worker ID MUST change for this to work.

C-Love (changes requested):
Additional tweaks needed.

actions/transfer-case.vbs
MiKaylaH22: moved to line 188 after we are done in REPT/USER

actions/transfer-case.vbs
MiKaylaH22: removed

actions/transfer-case.vbs
MiKaylaH22: updated and moved dialog to correct

actions/transfer-case.vbs
MiKaylaH22: this was also corrected with another issue

actions/transfer-case.vbs
MiKaylaH22: removed the forms to AREP

actions/transfer-case.vbs
MiKaylaH22: Removed unneeded lines of code

actions/transfer-case.vbs
MiKaylaH22: removed transfer_case because it was a superfluous variable

actions/transfer-case.vbs
MiKaylaH22: unclunck it

actions/transfer-case.vbs
C-Love: This doesn't appear resolved. Line 244 and 245 define the variables for the cash CFR and line 249  and 250 define the variables for the HC CFR, but lines 341 and 349 refer to a non-program-specific CFR variable, which are not repeated anywhere else in the script. 

line 182 and 347 refer to the county of financial responsibility but that variable doesn't match with what is defined in the dialog on line 243. 

the lines 384-400 seem to refer to the correct variables that are defined within the dialog, but the CASE:NOTE functionality won't match this information entry or action detail.

C-Love: Looking at this closer - I think the variable for 'cash_status' isn't defined anywhere.

MiKaylaH22: IF ga_case = TRUE or msa_case = TRUE or mfip_case = TRUE or dwp_case = TRUE or grh_case = TRUE THEN

C-Love (approved):
BAM

MiKaylaH22:
Updated the unmatched variables

Issue #807, opened 2022-04-14
NOTES - Interview - check if the PDF exists RIGHT before trying to open


C-Love:
Sometimes people get a Script Error when the PDF is trying to be opened. Likely a network issue. Add handling to make sure the PDF file path is found AFTER the message so there is no delay and message out if it appears that the path is gone.

C-Love:
This is done, no new errors reported.

PR #810, opened 2022-04-15
#807 NOTES - Interview - making sure the pdf is there before opening it

C-Love:
The script now does a check for the PDF file path right before trying to open it to reduce the appearance of a script error if the document has been moved or a network connection prevents access. 

If the document could not be found, the end message will read:
image


IlseFerris (approved):
I approve thee! 

Side bar: I created a local function in ADMIN - EXPEDITED REVIEW (lines 80 - 86) to support checking if a file exists. Let's talk about that at our next meeting.

Issue #667, opened 2021-12-07
ADMIN - On Demand Applications - Next Gen Phase


C-Love:
Changing the internal process within QI of handling the Applications of the On Demand list.

C-Love:
Need to make sure to include STATS output. #625

C-Love:
Updates needed now:
 - Add a column to the Daily Worklist Template for 'ECF Document Accepted', with droplist - 'Yes', 'Not Needed', 'Tomorrow'
 - Add automation to Work Assignment Completed to read counts 
 1. Reviewed - read from Action Taken column
 2. Denials - read for a Denial NOTE and check for I - make an alert of ones that mismatched
 3. PROG Updated - no automation
 4. APPT NOTC sent - read CASE NOTEs for the APPT NOTC NOTE
 5. NOMIs sent - read CASE NOTEs for the NOMI NOTE
 6. Email Correction sent - read from the email column in the worklist 
 7. Cases with NOTE added - read CASE NOTEs
 8. NEW - add one for Cases with ECF Docs Accepted 
 9. NEW - Does this case have a second application date - not fillable
 10. NEW - NO PRIV Access - not fillable
 - Create a Report List for Jen of information from Work Assignment Completed - in the archive. Include all of the counts and a hyperlink to the archived file for that day.
 - Add second application date column to the Worklist.

C-Love:
We have the SQL table in place to replace the BOBI data. 
We should run a trial during the Projects and Issues meeting to ensure IF and MH have access. 
If there are permissions issues we will contact BI - this will inform next steps and planning.

C-Love:
Remove BZST from the email reports. Make sure QI leadership has the emails they need.

C-Love:
On Demand is IN SQL AND WORKING AS A DASHBOARD!!!!!

DONE!!!!!!!!!!!!

PR #812, opened 2022-04-15
#667 ADMIN - Work Assignment Completed - Remove Correction Emails

C-Love:
only commenting out because we don't want to lose this until we know our final process


IlseFerris (approved):
Looks good. Just commenting out for now make sense. I approve thee!

Issue #468, opened 2021-05-20
Update all On Demand Notices for In Person Return


C-Love:
We are going to have to put back the in person locations when we start to go back in person. this will be a 'Phased In' approach to reopen sites.

We will also need to update the notice text block about the email document submission.

1. Do we want to update the notices on a 'phased in' approach or just once they are all open? This is a question for PM level or above)
2. What do we want to update the format to?

Current Notice -  this bulk of text is in ALL of our On Demand Notices:
    You now have an option to use an email to return documents           
    to Hennepin County. Write the case number and full name              
    associated with the case in the body of the email. Only the          
    following types are accepted PNG, JPG, TIFF, DOC, PDF, and           
    HTML. You will not receive confirmation of receipt or                
    failure. To obtain information about your case please                
    contact your worker. EMAIL: hhsews@hennepin.us                       

Possible new notice:
    You now have an option to return documents to Hennepin 
    County by email. Write the case number and full name              
    associated with the case in the body of the email. Only use          
    file types; PNG, JPG, TIFF, DOC, PDF, or HTML.          
    No reply or confirmation from this email.                
         *****    EMAIL: hhsews@hennepin.us    *****                       

Will need follow up.

- [x] On Demand Applicatiotns
- [x] Application Received
- [x] On Demand Individual Notices
- [x] Review Report

C-Love:
Updates to be made for the new format of our in person service:

- In Person option is in the form of digital access spaces.
- Interviews and case communication happens via phone
- virtual drop box is being phased out
- InfoKeep and MNbenefits needs to be added as options

These new notes will need to balance providing information about where to go and managing expectations of what is available in these spaces.

Planned to be in effect for May 2022.

C-Love:
All notices are now being sent in English. We need to decide if we want to pursue translation.

C-Love:
All updates are done. NOMIs have the website hennepin.us/economic-supports information included. There is no place to fit this in the Appt Notices unless we get a smaller url. 

At this time, we are closing this without adding this website to Appt Notices. If we get additional direction or information, we can revisit these notices. 

Closing this issue.

PR #819, opened 2022-04-22
#468 On Demand Renewal Notices updates for In Person Return

C-Love:
Updates to ADMIN - Review Report and ADMIN - Individual Recert Notices to include information on arriving in person to a digital access space if needed. 

This is the recert phase, which will be in effect starting 4/25/22. Another update to address Appplications will happen for 5/1/22

@IlseFerris - if you have a chance to review today or this weekend that would be great. If not, no worries. I am going to have to pull it in on Monday regardless. I will be the user of the script on Monday - so it should be fine. If I break it I can fix it then.


IlseFerris (approved):
This matches the verbiage we agreed on in email. I approve thee!

Issue #815, opened 2022-04-20
Basket Updates in Scripts for LTC+ and Housing Support Updates Going Live 04/25/2022


IlseFerris:
See the [Article](https://hennepin.sharepoint.com/teams/hs-economic-supports-hub/SitePages/Exciting-changes-coming-to-ADS-and-Housing-Supports-(GRH).aspx?xsdata=MDV8MDF8fDUzYzMwNzg1MmVjYjQzYjljYTZkMDhkYTIyZmNlN2VlfDhhZWZkZjlmODc4MDQ2YmY4ZmI3NGM5MjQ2NTNhOGJlfDB8MHw2Mzc4NjA3NjY4NDQ2MzU3NTd8R29vZHxWR1ZoYlhOVFpXTjFjbWwwZVZObGNuWnBZMlY4ZXlKV0lqb2lNQzR3TGpBd01EQWlMQ0pRSWpvaVYybHVNeklpTENKQlRpSTZJazkwYUdWeUlpd2lWMVFpT2pFeGZRPT18MXxNVGs2TnpKaU1XRmlZbUV3TW1FME5EVXlOVGcxTnpjMk5EbGtaVFkyTVRRNVpEUkFkR2h5WldGa0xuWXl8fA%3D%3D&sdata=SE9tWlJpSHdTc1FiR1gwMDRiYXF0WEc4OFJCZ1lqR3lkTG1qelFTVm83Zz0%3D&ovuser=8aefdf9f-8780-46bf-8fb7-4c924653a8be%2CIlse.Ferris%40hennepin.us&OR=Teams-HL&CT=1650482642538&params=eyJBcHBOYW1lIjoiVGVhbXMtRGVza3RvcCIsIkFwcFZlcnNpb24iOiIyNy8yMjAzMDcwMTYxMCJ9) posted on 04/12/2022 in Hot Topics about updates to LTC+ and Housing Supports change that go live Monday, April 25, 2022. 

Changes have been reviewed with and confirmed by project sponsor, MD. Spreadsheet with information held on shared drive. Specific baskets will be changing to either Housing Supports or LTC+, and the following scripts will need to be updated to support this work, and the associated links to the Q-Flow system.
- [x] ADMIN - FIND Q-FLOW POPULATION
- [x] NOTES - CLIENT CONTACT
- [x] TEST SCRIPTS - ILSE - GET BASKET NUMBER

[...]

Issue #747, opened 2022-03-07
Idea - Create document on Sharepoint to track all the basket logic we use


C-Love:
This is not the best place for the agency to keep it. 

We do need a way to have all of our basket logic/determinations in a spot we can find. Could we create an Excel document or Word document with the relevant basket details that we use in scripts?

IlseFerris:
We are not the product owners of this. Closing as won't fix.

PR #820, opened 2022-04-22
#815 ltc+ housing support basket updates

IlseFerris:
ADS baskets that do not have an identified population at this time have been commented out. Once these are finalized we should talk about a reserved variable for suggested_poptulation.


C-Love:
We might be able to incorporate suggested population and other basket logic in #747 - worth looking into. Maybe make it less Sharepoint focused.

C-Love (approved):
Looks solid - I did not check each basket individually - but the code looks solid.

Issue #814, opened 2022-04-19
MAXIS IVES DAIL Message Changes - Affects to DAIL Scrubber and Specific DAIL Scripts


IlseFerris:
Per[ SIR Announcement](https://www.dhssir.cty.dhs.state.mn.us/MAXIS/Lists/MAXIS%20Announcements/DispForm.aspx?ID=7610&RootFolder=%2FMAXIS%2FLists%2FMAXIS%20Announcements&Source=https%3A%2F%2Fwww%2Edhssir%2Ecty%2Edhs%2Estate%2Emn%2Eus%2FPages%2FDefault%2Easpx) dated 04/15/22 IEVS DAIL messages will no longer display participants' SSN and/or Name. MAXIS Ref Nbr will replace the participants' SSN and/or Name.

Scripts will need to be able to support both types of messages and the information they contain. Based on initial review, the following scripts need to be updated: 

DAIL - DAIL SCRUBBER: Updates need to read variables for old and new messages
- [x] dail/bndx-scrubber.vbs
- [x] dail/new-hire-ndnh.vbs
- [x] dail/new-hire.vbs
- [x] dail/sdx-info-has-been-stored.vbs
- [x] dail/tpqy-response.vbs 

DAIL - NEW HIRE
- [x] Update code to support finding the SSN or Member Number. Code will need to have two paths to get to the member information. 

DAIL - NEW HIRE NDNH
- [x] Update code to support finding the SSN or Member Number. Code will need to have two paths to get to the member information. 
- [x] Ensure there is no impact to interface when resolving the INFC message - SSN information will be needed to clear the interface portion of the message. 

DAIL - BNDX SCRUBBER
- [x] Update code to support finding the SSN or Member Number. 

DAIL - TPQY RESPONSE
- [x] Update code to support finding the PMI for the member from the MEMB panel to successfully navigate to the TPQY. 

DAIL - WAGE MATCH SCRUBBER
- [x] Update code to support finding the SSN for the member from the MEMB panel to successfully navigate to the INFC interface panels.

IlseFerris:
Whoops, that should not be closed.

IlseFerris:
MEMB 00 messages are appearing in HIRE matches. Emailed DHS BA about how to clear these on 05/12/22 as she is already working on updates to TE.02.08.142 per my request.

IlseFerris:
No issues have been reported by staff. Closing issue.

PR #831, opened 2022-05-04
#814 dail ievs bug fix

IlseFerris:
There was a bug in NDNH for gathering the member #/SSN if using the INFC portion of the script. Changed code to use the same code to gather this information and then go into If...statements for the specific processes (HIRE or INFC). 

Also updated BlueZone string entries to upper case, just in case.


C-Love (approved):
Looks good. There are SO MANY lower case letters that were in here!

Issue #486, opened 2021-06-02
Review ALL Dialog titles for more detail about what is happening/which script is running


C-Love:
We have a lot of generic dialog titles... or they just say 'Dialog' 

It would be good to have some detail in these dialogs. 
This is very true for the first 'Case Number' Dialog - also screen shots would be more helpful because we would know WHICH SCRIPT. 

This is also relevant in #125 to create a function to start the script - we want the script name in the dialog.

IlseFerris:
#125 will support new scripts (get this script started function). To complete this issue CLOSING ISSUE DOCUMENTATION will be updated to include checking the dialog name for information and space.

PR #833, opened 2022-05-05
#486 add Henn Co website to NOMIs 

C-Love:
Update to the following scripts:
ADMIN - Review Report
ADMIN - On Demand Waiver Applications
ADMIN - Individual NOMI
ADMIN - Individual Recertification Notices

To add the line
" More detail can be found at hennepin.us/economic-supports" 
after the address information in the notice. 

This will not be added to Appointment Notices at this time since there is not space to add an additional line without taking something away. 

I plan to pull this in on Monday 5-9-22. Please review before that time.


IlseFerris (approved):
Didn't test since it's just one line of code in each script. I don't see any thing concerning code-wise. I approve thee!

PR #834, opened 2022-05-07
#814 DAIL - NEW HIRE NDNH: Bug for INFC portion

IlseFerris:
If the SSN is known and we are just clearing the INFC we don't need to go to STAT at all. Added functionality to support if going to stat is necessary or not.

Successfully cleared old NDNH message in INFC, and tested new NDNH message in INFC (but they won't clear due to the known MAXIS issue).


MiKaylaH22 (approved):
Reviewed and ran script on a couple cases. Looks great, I like the agreement for INFC issues.

Issue #393, opened 2021-03-17
ACTIONS - TRANSFER CASE: Should this case note in all situations?


IlseFerris:
Transferring in-county is an internal process. As we are evaluating case notes and their functions we should review when a transfer is created in a universal caseload model.

MiKaylaH22:
Error reported: "CURRENT FIN RESP CO MUST BE BLANK IF EXCLUDED TIME CASE (Y/N) IS N"

MiKaylaH22:
testing workers reported they prefer the script to open the transfer case use form automatically but for now we are leaving the script as is. Can close per Ilse

PR #837, opened 2022-05-10
#393 ACTIONS TRANSFER CASE

MiKaylaH22:
adding a back-to-self and maxis background check to ensure transfer is complete, thinking we may need a script run lowdown in case the error is occurring at transfer

### DIFF
--- a/actions/transfer-case.vbs
+++ b/actions/transfer-case.vbs
@@ -86,6 +86,7 @@ function view_poli_temp(temp_one, temp_two, temp_three, temp_four)
 	EMWriteScreen "X", 6, 4
 	transmit
 end function
+
 '--------------------------------------------------------------------------------The script
 EMConnect ""                                        'Connecting to BlueZone
 CALL MAXIS_case_number_finder(MAXIS_case_number)    'Grabbing the CASE Number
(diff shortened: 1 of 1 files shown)


IlseFerris (commented):
Changes added to this pull are approved. I added comments on a couple of other areas in the code to address. 

Lines 205 - 210: This handling should be in the user dialog vs. letting the worker transfer the case to themselves. 

Lines 116 - 156: active_worker_found is defaulted to true and then will loop until true. Recommending to default to false and passing if found and active to true.

actions/transfer-case.vbs
```diff
@@ -139,8 +140,12 @@ DO
 
     '----------Checks that the worker or agency is valid---------- 'must find user information before transferring to account for privileged cases.
     IF transfer_to_worker = "X127CCL" THEN script_end_procedure("This case is will be transferred via an automated script after being closed for 4 months, the script will now end.")
-
+    IF worker_number = transfer_to_worker THEN          'If they match, cancel the transfer and save the information about the 'failure'
+        end_msg = "This case is already in the requested worker's number."
```
IlseFerris, line 144: Line 144 can be removed since the message is already in the script end procedure on line 145

IlseFerris (changes requested):
Just the one little thing and you should be good to go.

IlseFerris (approved):
I approve thee!

IlseFerris (approved):
I approve thee!

Issue #836, opened 2022-05-10
Update FuncLib - autofill_editbox_from_MAXIS - ABPS section needs a trim


C-Love:
The variable ABPS_current needs a trim right before the 'split'.

If there is a space entered at the beginning of the last name, split will create a null for one of the array items. This caseus a problem when trying to do string things to it later in the function.

Issue #829, opened 2022-05-04
Update FuncLib - navigate_to_MAXIS_screen - have the function read if we are already ON the requested screen


C-Love:
To prevent unnecessary moving, update this function to read if we are on the screen that has been entered. 

For example:
If requested to go to CASE/CURR, before any other navigation happens, look on the current screen to see what panel we are on and if already in the right place - don't even move. 

Thoughts here is that we only do the NON-PRIV-CHECK version since the priv check one should always move. What does everyone else think? @IlseFerris & @MiKaylaH22

C-Love:
@IlseFerris the function 'navigate_to_MAXIS_screen' does not check for if we are in the correct footer month. So If we are in STAT already and we want to go to 'STAT', 'MEMB' - it just goes to MEMB without confirming the footer month. It DOES confirm the MAXIS_case_number. Do we want to include that in this update OR should we look at a new issue for that?

IlseFerris:
@C-Love how much time/evaluation would you need to add it to this issue?

C-Love:
I think it would be minimal. I would keep the restriction to STAT and ELIG only. It's mostly a matter of ironing out the logic.

IlseFerris:
If it's minimal, then go ahead and include it in this issue. Thank you @C-Love!

C-Love:
This seems to be working correctly.

Issue #828, opened 2022-05-04
Update FuncLib determine_prog_case_status function to navigate differently if on the DAIL


C-Love:
Add handling at the beginning of the function to read if we are on the DAIL. If we are, have an if statement that will navigate using 'H' instead of navigate_to_MAXIS_screen so that we can retain the tie to the DAIL when this function is called in a DAIL Scrubber script.

This will not require an additional parameter since we can programmatically read if we are on the DAIL before any navigating functionality happens within the function.

We can also remember that we were on the DAIL and 'PF3' back to the dail once information is gathered.

PR #843, opened 2022-05-16
#836, #829, #828 Functions Library - Small bugs and enhancements

C-Love:
This pull resolves three issues:

#836 - Autofill_editbox_from_MAXIS - when reading ABPS, if there is a space before the last name, a null item is created when split into an array of details in the function. Adding a trim to the variable before splitting will resolve this issue. 

#828 - determine_program_and_cases_status - changing the navigation to CASE/CURR to handle differently if starting from the DAIL.

#829 - navigate_to_MAXIS_screen - check to see if we are already at the right screen before having the script move.

Ready for review.


MiKaylaH22 (approved):

IlseFerris (approved):
Reviewed code, looks good. I approve thee!

PR #844, opened 2022-05-16
#814 DAIL - BNDX SCRUBBER: added handling for both SSN_present TRUE or FALSE

IlseFerris:
Tested on both types of messages. We don't have any in county but I can provide a caseload out of county to test.


MiKaylaH22 (approved):
looks good

Issue #855, opened 2022-05-24
ACCESS_ADDR_PANEL Function not entering line 2 of Mailing address on WRITE


C-Love:
Bug in the function 'access_ADDR_panel' where if a second line for mailing address in entered, it does not get added to the ADDR panel correctly. 

This relates to an issue reported in NOTES - Returned Mail on #427. The bug will be resolved with this issue.

C-Love:
DONE - no errors reported

Issue #427, opened 2021-04-08
NOTES-RETURNED MAIL update PROG 


MiKaylaH22:
- [x]  We need to read PROG to determine for CASH cases if the active program is listed on the Cash 1 line or Cash 2 line in PROG because that is the line of the PACT panel that needs to be updated.
- [x]  Update closing message to point users to WCOM supports in NOTICES - ADD WCOM.
- [x] on the pop up box if you enter apt in on the 2nd line it does not populate on the addr panel but it does show up in the case note - i did enter in the unit # manually to make sure all mailings get to HH correctly

C-Love:
We need to read PROG to determine for CASH cases if the active program is listed on the Cash 1 line or Cash 2 line in PROG because that is the line of the PACT panel that needs to be updated.

MiKaylaH22:
to ensure process integrity we will be creating a dialog that asks the worker to confirm that they are entering a pact panel and agree to do so. also add the poli/temp button. review to remove success message 791 and 743 review verbiage

IlseFerris:
@MiKaylaH22 when you're making this update, you won't need to have the local function view_poli_temp in your code. Issue #839 put the function into the FuncLib.

MiKaylaH22:
No issues reported but reviewing usage to ensure all issues are addressed

PR #857, opened 2022-05-24
#855 & #427 access_ADDR_panel bug in updating an addr with 2 lines

C-Love:
This pull will resolve the acces_ADDR_panel function bug and also align NOTES - Return Mail to use this function well.


IlseFerris (approved):
Code looks good to me. I approve thee!

PR #859, opened 2022-05-25
814 closing project documentation

IlseFerris:
See notes in the commits


dail/new-hire-ndnh.vbs
```diff
@@ -303,7 +306,7 @@ IF match_answer_droplist = "NO-RUN NEW HIRE" THEN
 	'-----------------------------------------------------------------------------------------CASENOTE
     Call navigate_to_MAXIS_screen("CASE", "NOTE")
```
C-Love, line 307: Is there a reason we aren't using 'start_a_blank_case_note' - using the navigate_to function can already lose the tie to the dail.

IlseFerris: Do you think it makes that much of a difference since start_a_new_case_note also uses navigate_to_MAXIS_screen?

C-Love (approved):

PR #869, opened 2022-06-01
#393 ACTIONS - TRANSFER CASE - bug fix

MiKaylaH22:
Error reported: "CURRENT FIN RESP CO MUST BE BLANK IF EXCLUDED TIME CASE (Y/N) IS N"
blanked out all of the fields if excluded time is no

### DIFF
--- a/actions/transfer-case.vbs
+++ b/actions/transfer-case.vbs
@@ -201,18 +201,14 @@ If transfer_out_of_county = False THEN      'If a transfer_to_worker was entered
 ELSE 'this means out of county is TRUE '
     CALL determine_program_and_case_status_from_CASE_CURR(case_active, case_pending, case_rein, family_cash_case, mfip_case, dwp_case, adult_cash_case, ga_case, msa_case, grh_case, snap_case, ma_case, msp_case, emer_case, unknown_cash_pending, unknown_hc_pending, ga_status, msa_status, mfip_status, dwp_status, grh_status, snap_status, ma_status, msp_status, msp_type, emer_status, emer_type, case_status, list_active_programs, list_pending_programs)
 
-    'Using move date to determine CRF change date.
-
-    client_move_date = date 'this doesnt enter into the dialog and I am unsre why'
-    excluded_time_begin_date = date
+    client_move_date = date & ""
+    excluded_time_begin_date = date & ""
     cfr_month = CM_plus_2_mo
 	cfr_year = CM_yr
 
     hc_cfr_month = cfr_month' this variable may be changed by the worker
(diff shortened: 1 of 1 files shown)


MiKaylaH22:
Discovery: blank out the current county of fin resp and enter the transferring county code (current) - will need to know if maxis overwrites

actions/transfer-case.vbs
```diff
@@ -274,13 +273,13 @@ ELSE 'this means out of county is TRUE '
...
 
     'Using move date to determine CRF change date.
-     cfr_date = dateadd("M", 1, client_move_date)
-     cfr_date = datepart("M", cfr_date) & "/01/" & datepart("YYYY", cfr_date)
-     cfr_date = dateadd("M", 2, cfr_date)
-     cfr_month = datepart("M", cfr_date)
-     IF len(cfr_month) <> 2 THEN cfr_month = "0" & cfr_month
-     cfr_year = datepart("YYYY", cfr_date)
-     cfr_year = right(cfr_year, 2)
+    cfr_date = dateadd("M", 1, client_move_date)
```
C-Love, line 276: This should be moved up to before the dialog and used to set the variables used in the dialog

actions/transfer-case.vbs
```diff
@@ -359,45 +358,30 @@ ELSE 'this means out of county is TRUE '
...
-            EMWriteScreen hc_cfr_year, 14, 59
-        ELSE
-            IF ma_case = True THEN
-                EMWriteScreen hc_cfr, 14, 39
-                EMWriteScreen cfr_month, 14, 53
-                EMWriteScreen cfr_year, 14, 59
-            END IF
+        IF (ga_case = TRUE or msa_case = TRUE or mfip_case = TRUE or dwp_case = TRUE or grh_case = TRUE) THEN 'previously we read PROG for cash one and cash two programs unsure if this is necessary'
+            EMWriteScreen cash_cfr, 11, 39
+            EMWriteScreen cfr_month, 11, 53
```
C-Love, line 374: The month and year variables don't match the dialog variables here.

MiKaylaH22: got it, i had to keep the hc_cfr_month variable incase it is different from the cash per the POLI

actions/transfer-case.vbs
MiKaylaH22: this cant be moved before the dialog cfr_date = dateadd("M", 1, client_move_date) but I can add date stuff before the dialog I will follow up with you tomorrow - since dates

C-Love (approved):
This should work!!!!

PR #879, opened 2022-06-17
873 dail remedial care updates

IlseFerris:
Updated Remedial care amounts for July 2022. Unable to fully test due to lack of remedial care messages currently


MiKaylaH22 (approved):
Looks like it will work

PR #881, opened 2022-06-21
880 paris dail bug

IlseFerris:
See commits - 1st for no SSN supports. 2nd for critical fixes/updates. 3rd for Closing issue documentation. 

I did not clear any messages, but I got the script to work all the way through to just before updating and the case note. Let's get some DEU folks lined up right away in the AM to test this.

### DIFF
--- a/deu/paris-match-cleared.vbs
+++ b/deu/paris-match-cleared.vbs
@@ -58,15 +58,13 @@ changelog_display
 'END CHANGELOG BLOCK =======================================================================================================
 '---------------------------------------------------------------------THE SCRIPT
 EMConnect ""
-Call Check_for_MAXIS(False )
+Call Check_for_MAXIS(False)
 
 EMReadscreen dail_check, 4, 2, 48
 IF dail_check <> "DAIL" THEN script_end_procedure("You are not in your dail. This script will stop.")
 
-'TYPES A "T" TO BRING THE SELECTED MESSAGE TO THE TOP
-EMSendKey "T"
+EMSendKey "T" 'TYPES A "T" TO BRING THE SELECTED MESSAGE TO THE TOP
 TRANSMIT
-
 EMReadScreen DAIL_message, 4, 6, 6 'read the DAIL msg'
 IF DAIL_message <> "PARI" THEN script_end_procedure("This is not a Paris match. Please select a Paris match, and run the script again.")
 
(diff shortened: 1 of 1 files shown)


deu/paris-match-cleared.vbs
```diff
@@ -90,26 +112,26 @@ DO
...
+	IF INTM_match_status = "RV" THEN
+        row = row + 1 'skipping verified cases
+    Else
+        INTM_info_confirmation = MsgBox("Press YES to confirm this is the match you wish to act on." & vbNewLine & "For the next match, press NO." & vbNewLine & vbNewLine & _
         "   " & INTM_period, vbYesNoCancel, "Please confirm this match")
-		IF INTM_info_confirmation = vbNo THEN
+	    IF INTM_info_confirmation = vbNo THEN
             row = row + 1
-            'msgbox "row: " & row
             IF row = 18 THEN
```
C-Love, line 123: This might be needed in the row = row + 1 on line 117, it could hit the end of the list and need to page down. It should be safe to move outside of the ELSE because if vbYes is hit - the row should not increment and so won't hit 18.

deu/paris-match-cleared.vbs
```diff
@@ -299,73 +319,63 @@ IF send_notice_checkbox = CHECKED THEN
...
-	IF sent_date <> "" then MsgBox("A difference notice was sent on " & sent_date & ". The script will now navigate to clear the PARIS match.")
+    		err_msg = ""
+    		Dialog Dialog1
+    		cancel_without_confirmation
+    		IF bene_other_state = "Select One:" THEN err_msg = err_msg & vbNewLine & "* Is the client accessing benefits in other state?"
+    		IF contact_other_state = "Select One:" THEN err_msg = err_msg & vbNewLine & "* Did you contact the other state?"
+    		IF fraud_referral = "Select One:" THEN err_msg = err_msg & vbnewline & "* You must select a fraud referral entry."
+    		IF err_msg <> "" THEN MsgBox "*** NOTICE!!! ***" & vbNewLine & err_msg & vbNewLine
+    	LOOP UNTIL err_msg = ""
+    	CALL check_for_password(are_we_passworded_out)
```
C-Love, line 333: Looks like this previous code had issues if there was a transmit - does that not pose a concern here anymore?

deu/paris-match-cleared.vbs
```diff
@@ -410,18 +426,7 @@ ELSEIF clear_action_checkbox = CHECKED or notice_sent = "Y" THEN
...
-    	IF MAXIS_check <> "MAXIS"  and MAXIS_check <> "AXIS " THEN
-    		IF end_script = True THEN
-    			script_end_procedure("You Do not appear to be in MAXIS. You may be passworded out. Please check your MAXIS screen and try again.")
-    		ELSE
-    			warning_box = MsgBox("You Do not appear to be in MAXIS. You may be passworded out. Please check your MAXIS screen and try again, or press ""cancel"" to exit the script.", vbOKCancel)
-    			IF warning_box = vbCancel THEN stopscript
-    		END IF
-    	END IF
-		CALL check_for_password_without_transmit(are_we_passworded_out)
+		CALL check_for_password(are_we_passworded_out)
```
C-Love, line 429: Same question here on the transmit

C-Love (changes requested):
The row incrementor comment could cause an issue. the other two comments might be a non-issue - I am not familiar enough with the MAXIS flow here to know - just bringing awareness to review.

deu/paris-match-cleared.vbs
IlseFerris: Nope, this should stay on the same panel since it's not in edit mode and transmitting won't bring the user to another panel.

C-Love (approved):
Looks good within the scope of the issue.

deu/paris-match-cleared.vbs
C-Love: The update you made should do what is needed

Issue #882, opened 2022-06-22
UTILITIES-UC VERIFICATION REQUEST - uncheck all HH members


MiKaylaH22:
Reported: Users reported a need to uncheck members
Recommendation: Uncheck all HH members because it is easier to select a few

MiKaylaH22:
No issues reported

PR #883, opened 2022-06-23
#882 UTILITIES-UC VERIFICATION REQUEST

MiKaylaH22:
uncheck all HH members, had to pull in a custom function and uncheck it that way, discussed with Ilse updating function


IlseFerris (approved):
I approve thee!

Issue #493, opened 2021-06-10
CREATE A FUNCTION FOR NON-DISCLOSURE AGREEMENT REQUIRED FOR ACCESS


MiKaylaH22:
PRISM
TPQY
INFC

MiKaylaH22:
[3:46 PM] Kyle C Nelson
'Wait a moment, then check to see if the MAXIS authorized user agreement screen is being displayed. If it is, ask the user to click OK once they have finished reading it.
EMWaitReady 10, 250
EMReadScreen check_for_warning, 7, 3, 3
IF check_for_warning = "WARNING" THEN
MsgBox "Have you finished reading the MAXIS authorized user agreement text?" & vbcr & vbcr & "Click OK to continue.", vbExclamation, "MAXIS Authorized User Agreement"
transmit
END IF

IlseFerris:
@MiKaylaH22 there is some handling for the agreement in DAIL - BNDX SCRUBBER you might want to check out.

MiKaylaH22:
- [ ] actions/send-sves.vbs
- [ ] actions/check-edrs.vbs
- [ ] dail/es-referral-missing.vbs
- [ ] dail/cola-sves-response.vbs
- [ ] dail/tpqy-response.vbs
- [ ] dail/bndx-scrubber.vbs
- [ ] Test scripts/Ilse/Other/BULK - SSI-RSDI UPDATER.vbs
- [ ] deu/paris-match-cleared.vbs -already done
- [ ] deu/match-cleared.vbs
- [ ] deu/atr-received.vbs
- [ ] deu/paris-match-cleared-CC-claim-entered.vbs
- [ ] deu/overpayment-claim-entered.vbs
- [ ] deu/bulk-match-cleared.vbs

MiKaylaH22:
actions/send-sves.vbs - per our standardization -  removed radio buttons (SVES TE02.12.13)

MiKaylaH22:
The following scripts are excluded from handling
admin/training-case-creator.vbs - does not require
admin/send-cbo-manual-referrals.vbs -WF1M doesn't  require
notices/snap-e-and-t-letter.vbs - does not require
notices/dwp-es-referral.vbs - does not require
dail/dail-scrubber.vbs - does not engage INFC 
dail/new-hire.vbs hire - does not require
dail/financial-orientation-missing.vbs - WORK does not require

MiKaylaH22:
Nothing has been reported

Issue #498, opened 2021-06-16
DEU BULK  MATCH CLEARED - bug not clearing all cases 


MiKaylaH22:
ASSOCIATED #216 

- [x] ensuring that a case note is not made if the match is not cured 
- [x] Multiple programs - skip logic add manual clearing for 2 programs on the excel sheet
- [x] If cleared messages "" or not applicable = False
- [x] Case not skipping if match not cleared.
- [x] Don't work through the DAIL
- [x] change the dialog box to ensure that a cleared method is entered

MiKaylaH22:
we will be retiring the scripting tool, DEU – BULK MATCH CLEARED. Some of the reasons for this include but are not limited to:
•	A reduction of matches of Health Care matches due to the Health Care Exchange (METS)
•	The need for the BlueZone Scripts project to reduce redundancies in our automation supports for future conversion to a new programming language. 
•	Reduce confusion for DEU staff who are clearing in BULK and/or clearing matches on a case-by-case basis.

MiKaylaH22:
BUG reported: There were a few that did not clear.   I have attached a file with 2 cases from each HSR that did not clear.   I can not figure out why the matches are not clearing.   I am guessing it has to do with change from ssn to member #.   Maybe the relationship is why??   Can you see what you can figure out?    There is a total of about 400 matches that will need to be run still using the Bulk Cleared script. 
Requested that she manually check these cases

MiKaylaH22:
Need to add new Supervisors users ID as previous will be done mid July.

MiKaylaH22:
nothing reported

PR #889, opened 2022-07-01
#493 handling for non disclosure agreement

MiKaylaH22:
Mutli script review for added handling


MiKaylaH22:
Had to revert #498 BULK MATCH CLEARED - adding the new HSS login infomartion

IlseFerris (approved):
The Non-disclosure stuff looks good. DEU scripts are showing lots of changes, and I don't have cases to test. Those may need updates at some point outside of this issue. 

I approve thee!

Issue #891, opened 2022-07-02
ADMIN - 12 MONTH CONTACT: PRIV Handling Bug


IlseFerris:
Script works great until it hit a PRIV case. Updated PRIV handling.

PR #899, opened 2022-07-07
#891 ADMIN - 12 MONTH CONTACT: Fixed bug w/ Priv cases

IlseFerris:
MAXIS Background check was in a spot that was causing issues with the PRIV handling. Tested with rest of 07/22 list, and is working as intended.


MiKaylaH22 (approved):
this one was very difficult, but good job!

Issue #922, opened 2022-07-26
DEU- MATCH CLEARED - match is not clearing due to IULB not updating


MiKaylaH22:
Problem:  error on IULB "enter comments to explain actions taken" 
Resolution:  evaluate do loop on IULB to ensure match is cleared before case noting

Findings: "BO-Other" was commented out as the HC cases were not to be cleared BO during he pandemic. The pandemic is not ending until Oct currently so we will add handing to ensure HC cases are not cleared "BO".

MiKaylaH22:
Nothing reported

Issue #921, opened 2022-07-26
DEU - BULK - MATCH CLEARED - remove STAT/MISC handling for HC matches


MiKaylaH22:
Problem:  HC matches that have FS programs are coming up on STAT/MISC report 
Resolution:  remove STAT/MISC handling for all HC matches on BULK list

MiKaylaH22:
No issues reported some enhancements coming

PR #924, opened 2022-07-26
DEU - match cleared & bulk match cleared #921 & #922

MiKaylaH22:
921-Problem: HC matches that have FS programs are coming up on STAT/MISC report
Resolution: remove STAT/MISC handling for all HC matches on BULK list

922-Problem: error on IULB "enter comments to explain actions taken"
Resolution: evaluate do loop on IULB to ensure match is cleared before case noting
Findings: "BO-Other" was commented out as the HC cases were not to be cleared BO during he pandemic. The pandemic is not ending until Oct currently so we will add handing to ensure HC cases are not cleared "BO".
The Supervisor has decided to allow the workers to select BO for HC and the terminology on IULB will be No review due during the match period.


IlseFerris (approved):
I approve thee!

PR #932, opened 2022-08-10
931 notes health care transition verbiage

IlseFerris:


C-Love (approved):
All seems right - I don't see any issues. I did not test this, but I don't see why it would be a problem.

PR #933, opened 2022-08-10
914 bulk fmed messages

IlseFerris:
Instructions in this folder: T:\Eligibility Support\Mnsure\SSR Folder\SHELTER PROJECT\REPORTS Instructions\Laurie H Directions for Running Reports

Tested for 08/2022 PEPR messages without issues.


C-Love (approved):
Solid code. Looks good.  Did not run but this looks like a thing we are running so it should all be fine.

Issue #942, opened 2022-08-25
ADMIN - Review Report has the LONGEST LIST OF PROGRAMS EVER - MEMO was too big


C-Love:
Had to adjust format for a case that had MFIP, SNAP, MSA, GRH AT THE SAME TIME. 

The MEMO went over. Removed an asterisk - it's fine now.

C-Love:
This is fixed!

PR #943, opened 2022-08-25
#942 - ADMIN - Review Report - Bug with long programs list - MEMO too big

C-Love:
Reformat of MEMO for long list of programs. One character change


IlseFerris (commented):
I approve thee!

Issue #940, opened 2022-08-24
ADMIN - Expedited Determination Report - update functionality to support Expedited Daily HSS Assignments


C-Love:
Review verbiage and functionality to support the daily information reports to HSSs about still pending Expedited cases.

C-Love:
Script functionality is finished.

PR #950, opened 2022-09-02
#940 ADMIN - Exp Det Report - worker data incorporated

C-Love:
Script will create emails and NOT send yer and update the reports.


IlseFerris (approved):
I approve thee!

Issue #954, opened 2022-09-03
Update FuncLib's HEST Standards for FFY 2023


IlseFerris:
[SIR announcement](https://www.dhssir.cty.dhs.state.mn.us/MAXIS/Lists/MAXIS%20Announcements/DispForm.aspx?ID=7755&RootFolder=%2FMAXIS%2FLists%2FMAXIS%20Announcements&Source=https%3A%2F%2Fwww%2Edhssir%2Ecty%2Edhs%2Estate%2Emn%2Eus%2FPages%2FDefault%2Easpx) dated 09/01/2022 announced the HEST standards that go into effect for footer month of 10/22 or for Federal Fiscal Year 2023.

Full HEST amt: $586 (increase)
Electric amt: $185 (increase)
Phone amt: $55 (decrease)

PR #955, opened 2022-09-03
#954 FuncLib: Updated FFY 2023's HEST Standards

IlseFerris:


C-Love (changes requested):
I think the year updates aren't done here. It still reflects 2021 in the DateDiff comparisons.

IlseFerris:
> I think the year updates aren't done here. It still reflects 2021 in the DateDiff comparisons.

But don't you like how I updated EVERYTHING but the date in the DateDiff?

C-Love (approved):
That looks right!

Issue #937, opened 2022-08-19
Connect to SQL Table about PENDING and EXP Cases


C-Love:
Update scripts to add information gathered to script run to new table created by BI so that they can pull data about pending cases and expedited cases into the assignment lists for correct priority.

Date to source from:
 - NOTES - Application Received 
 - Expedited Determination Functionality

Need table created - work with BI Team.

C-Love:
We will be able to remove the Expedited Baskets from the process with this change. We want to remove the Expedited Basket suggestion once this process is removed.

PR #958, opened 2022-09-07
#937 Adding pending cases to a SQL Table for ES Workflow

C-Love:
This commit includes:
 - NOTES - Application Received to add any case that is pending SNAP, or pending Cash and SNAP is not ACTIVE. These are defaulted to being potentially Expedited
 - NOTES - Expedited Determination - Uses the expedited determination to update the SQL Table with the actual information (makking some NOT Expedited.
 - NOTES - CAF - Uses the expedited determination to update the SQL Table with the actual information (makking some NOT Expedited.

Still needed:
 - NOTES - Interview - add the same functionality as CAF and Exp Det - this is currently branched in another pull request and needs to be resolved.

Testing will be tricky with this one and will require we monitor closely to ensure no issues.

### DIFF
--- a/Test scripts/Casey/Utilities/delete-casey-entry-from-pending-table.vbs
+++ b/Test scripts/Casey/Utilities/delete-casey-entry-from-pending-table.vbs
@@ -1,0 +1,29 @@
+Const adOpenStatic = 3
+Const adLockOptimistic = 3
+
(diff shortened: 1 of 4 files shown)


IlseFerris (approved):
This is exactly what I expected when we were talking about this. I approve thee!

Issue #962, opened 2022-09-09
There is a blank CASE/NOTE line in Exp Det functionality for FACI


C-Love:
Discovered a line in the scripts with Expedited Determination functionality that will put a line in CASE/NOTE that looks like this:
"     - "

So we don't need that there. Remove line of code that has this random blank line

C-Love:
This is done.

PR #963, opened 2022-09-09
#962 & #937 Add data output to SQL for Interview, CNOTE bug in Interview, CAF, Exp Det

C-Love:


IlseFerris (approved):
Looks good. Yay interview. I approve thee!

Issue #938, opened 2022-08-24
Update Read_ADDR function to have the notes_on_address output a statement


C-Love:
Current format is that notes_on_address is "Address effective: DATE."

Change this to be "Address effective on DATE."

C-Love:
This text is done.

Issue #864, opened 2022-05-26
Move pulling the tester list to Global Variables


C-Love:
This will ensure that only people with T Drive access will try to pull the list of testers from T Drive if they have access.

This requires the confirm tester call to have a n IsArray Boolean to only call the function if the array exists.

This is to support collaboration and VPN issues by scriptwriters. This would be part of the support for #566

C-Love:
We need line 83 - 91 in the Global Variables (line 83 should also stay in the FuncLib 
```
t_drive = "\\hcgg.fr.co.hennepin.mn.us\lobroot\hsph\team"

'LOADING LIST OF TESTERS====================================================================================================
tester_list_URL = t_drive & "\Eligibility Support\Scripts\Script Files\COMPLETE LIST OF TESTERS.vbs"        'Opening the list of testers - which is saved locally for security
Set run_another_script_fso = CreateObject("Scripting.FileSystemObject")
Set fso_command = run_another_script_fso.OpenTextFile(tester_list_URL)
text_from_the_other_script = fso_command.ReadAll
fso_command.Close
Execute text_from_the_other_script
```

Then line 101 from the FuncLib needs logic.

C-Love:
We did not break everything and all the things appear to work.

PR #965, opened 2022-09-12
#938 & #864 Master Functions Library - Move the Tester Call to Global Variables and change notes_on_addr verbiage

C-Love:
This pull will require coordination of the change to the T Drive version of Global Variables as that is accessed locally.


IlseFerris (approved):
I approve thee!

Issue #944, opened 2022-08-26
NOTES - IMIG - EMA - add TIKL 


MiKaylaH22:
I use the IMIG – EMA script often and sometimes forget to put a tikl out for 45 days if the EMA application is incomplete.  Can we have the IMIG – EMA script tikl out for 45 day check or deny like the HCAPP script does if the EMA script is coded as incomplete for the ‘actions taken’ section?

PR #967, opened 2022-09-12
#944 NOTES - IMIG - EMA - add TIKL

MiKaylaH22:
- [x] add tikl out for 45 day check
- [x] also added PRIV handling


IlseFerris (approved):
I approve thee!

Issue #969, opened 2022-09-12
UTILITIES - Copy Panel to Word - create function to determine if a page break is needed


C-Love:
The same functionality is used a dozen times and it should be a function.

C-Love:
No errors reported.

Issue #735, opened 2022-03-03
UTILITIES - COPY PANELS TO WORD: Enhance to include the PIC


IlseFerris:
Request from Ann S. (appeals) to add PIC information to COPY PANELS TO WORD script functionality. This would include the PIC for all income based panels for SNAP and GRH. 

Initial idea on how to complete this request: Add question to each income based panel to see if user needs the PIC, and for which program(s).

C-Love:
No errors reported.

PR #970, opened 2022-09-12
#735 & #969 - UTILITIES - Copy Panels to Word - Output PIC and add a function for insert page break

C-Love:
Script is updated to support the output of PIC and Income Calculation information from STAT panels. 

Created a function to correctly format the document - adding page breaks where they are needed


IlseFerris (approved):
I approve thee!

Issue #952, opened 2022-09-03
ACTIONS - CHECK EDRS: Bug in dialog loop


IlseFerris:
Update needed for when a user presses POLI TEMP button is pushed in the main dialog. If the button is pressed, the bug occurs when the user is leaving the dialog. There is no way to get out of POLI TEMP.

PR #972, opened 2022-09-12
#952 ACTIONS - CHECK EDRS: Bug in dialog loop

MiKaylaH22:
Update needed for when a user presses POLI TEMP button is pushed in the main dialog. If the button is pressed, the bug occurs when the user is leaving the dialog. There is no way to get out of POLI TEMP.


IlseFerris (approved):
I approve thee!

Issue #909, opened 2022-07-18
Copy CASE NOTE to Word


C-Love:
Create a UTILITY Script that will pull the contents of a CASE NOTE into word documents. 

First stage is to have the script run in on a single note that we are already in. 

Second stage is to create a menu to select multiple NOTEs and put them in one page.

PR #973, opened 2022-09-12
#909 New Script - UTILITIES - Copy Case Note to Word

C-Love:


IlseFerris (changes requested):
Add category for UTILITIES menu and then good to go

PR #975, opened 2022-09-12
923 memb 00 support

IlseFerris:


C-Love (approved):
This all looks good.
Approval granted!
Couldn't test NDNH with no messages.

IlseFerris:
Yay! We'll have to keep this one in testing for a bit until we get some NDNH messages.

PR #989, opened 2022-09-26
#667 - Move On Demand Applications to using SQL instead of BOBI

C-Love:
This is functionality only to change the initial data pull to getting information from a SQL table with MAXIS data from the InfoStore. 

This change does NOT alter the experience of QI processing this worklist. The working excel and worklist are still managed in Excel.


IlseFerris (approved):
Looking good. I approve thee!

Issue #990, opened 2022-09-28
BUG on NOTES - Emergency  for EGA screening when Income Equals Shelter


C-Love:
If information is entered in the EGA screening such that the 70% of net income is equal to shelter expense, the script identifies these cases as NOT eligible for EGA BUT the msgbox about the screening doesn't list why it isn't eligible. 

Confirmed with PV and GY that if equal they should screen as eligible. 

This is a bug.

C-Love:
This is done, no issues reported.

PR #993, opened 2022-09-30
#990 NOTES - Emergency - EGA screening bug when amounts are equal

C-Love:
This is a quick fix with just a single character and a changelog. Full project documentation and such not completed as this is the quickest of fixes.


IlseFerris (approved):
I approve thee!

Issue #798, opened 2022-04-05
ACTIONS - MA EPD EI FIATer - Add support for budget periods that are not entirely MA EPD


C-Love:
Currently the ACTIONS - MA EPD EI FIATer can only FIAT if the entire budget is set up for every month to be EPD. 

Request has come in from KW to enhance this functionality to support FIAT if only a part of the 6 month budget is EPD. This is not currently a large scope for number of cases but was prior to COVID Waivers and will likely be a large issue when normal HC processing returns.

C-Love:
Moving to Q3 to align with HC Renewals resuming.

C-Love:
It's a RACE - who can do this first?!

IlseFerris:
I removed me from assignees. @C-Love you did all the work.

C-Love:
Work is done!

PR #995, opened 2022-09-30
#798 ACTIONS - MA EPD FIATer - allow partial MA EPD Budget cases to work 

C-Love:
Previously all of the months in the budget needed to be MA-EPD to work correctly in the FIATer.

Updated the functionality to be able to identify an EPD month individually in the budget and only use the EPD month for the average AND only FIAT the months that are coded EPD.

### DIFF
--- a/actions/ma-epd-ei-fiat.vbs
+++ b/actions/ma-epd-ei-fiat.vbs
@@ -1,9 +1,9 @@
 'Required for statistical purposes==========================================================================================
 name_of_script = "ACTIONS - MA-EPD EI FIAT.vbs"
 start_time = timer
-STATS_counter = 1                     	'sets the stats counter at one
-STATS_manualtime = 600                	'manual run time in seconds
-STATS_denomination = "C"       		'C is for each CASE
+STATS_counter = 0                     	'sets the stats counter at one
+STATS_manualtime = 100                	'manual run time in seconds
+STATS_denomination = "I"       		'I is for Item - this is each MONTH that is FIATed
 'END OF stats block=========================================================================================================
 
 'LOADING FUNCTIONS LIBRARY FROM GITHUB REPOSITORY===========================================================================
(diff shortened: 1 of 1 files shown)


actions/ma-epd-ei-fiat.vbs
```diff
@@ -834,83 +835,115 @@ EMReadScreen elig_type_check_fourth_month, 2, 12, 50
 EMReadScreen elig_type_check_fifth_month, 2, 12, 61
 EMReadScreen elig_type_check_sixth_month, 2, 12, 72
 
-'Script will end if not DP
-If elig_type_check_first_month <> "DP" or elig_type_check_second_month <> "DP" or elig_type_check_third_month <> "DP" or elig_type_check_fourth_month <> "DP" or elig_type_check_fifth_month <> "DP" or elig_type_check_sixth_month <> "DP" then script_end_procedure("Not all of the months of this case are MA-EPD. Process manually.")
+'There needs to be at least 1 month of MA-EPD Elig Results - the script will check each month and if all are NOT DP - the script will end.
+If elig_type_check_first_month <> "DP" and elig_type_check_second_month <> "DP" and elig_type_check_third_month <> "DP" and elig_type_check_fourth_month <> "DP" and elig_type_check_fifth_month <> "DP" and elig_type_check_sixth_month <> "DP" then script_end_procedure("Not all of the months of this case are MA-EPD. Process manually.")
```
IlseFerris, line 839: Script end procedure should be updated to say no MA-EPD exists since the update is to account for partial DP budgets.

IlseFerris (changes requested):
One quick update on the script end procedure, otherwise looks good. Did not test code, just visual review.

actions/ma-epd-ei-fiat.vbs
C-Love: Good call - updated.

IlseFerris (approved):
I approve thee!

Issue #703, opened 2022-02-02
Add email assignment to assignee option to ADMIN - WORK ASSIGNMENT FROM EXCEL


IlseFerris:
Jennifer Frey, QI supervisor, is requesting an update to this amazing scripting tool to provide an option to email the worker once the assignment is ready.

IlseFerris:
Jen loves this idea! Let's discuss at our next meeting how to make this happen.

C-Love:
Work is done - has not been used. If there is a bug, we will create a new issue.

PR #998, opened 2022-10-01
#703 ADMIN - Work Assignment from Excel

C-Love:
Adding the option to send an email of the assignments when completed. 

Additionally adds an option for selecting from QI members in a checkbox.


IlseFerris (approved):
I approve thee!

Issue #996, opened 2022-10-01
ADMIN - Review Report - restart functionality for End of Month processing


C-Love:
The Review Report failed on me during the End of Processing month run.

Since this functionality enters a CASE/NOTE, I did not want multiple NOTEs for the same thing. I had to slightly restructure the stats gathering (where the note happens) to be able to restart the functionality and have it pick up on the NOTE where it left off. 

This work is already done as there are time limits on these report runs.

C-Love:
This appears to be working.

Issue #949, opened 2022-09-01
ADMIN - Review Report - detail is coming in weird


C-Love:
This might be an access issue with the new ES-SS running the statistics and it may have to do with PRIV access. Basically dates look all weird. Need to review the cases where the information looks wrong in 08/22 or 09/22 REVW Report.

C-Love:
This appears to have not broken. Closing.

PR #1001, opened 2022-10-01
#949 & #996 ADMIN - MONT and Review Report - restart stats - stats looks weird in REVW

C-Love:
Some updates to make restarts possible and me sure we don't pull data from the wrong place in REVW.


IlseFerris (approved):
I approve thee! Visual review vs. testing, but nothing looks odd.

Issue #1004, opened 2022-10-04
ACTIONS - MA EPD FIATer - BUG - variable mismatch on member number information


C-Love:
There are 2 different variables being used for MA EPD FIATer which could cause some issues.

C-Love:
This is done.

PR #1005, opened 2022-10-04
#1004 ACTIONS - MA EPD FIATer - bug on memb numb variable

C-Love:
Not sure the impact here - I have marked it as critical because of the recent change and communication. 

We actually haven't heard any issues from this but the variable used for navigation and the variable entered in the dialog for the member number do not match.


IlseFerris (approved):
I approve thee!

PR #1007, opened 2022-10-06
1006 task based dail capture updates

IlseFerris:


C-Love (approved):
Looks good to me. 

(though I still maintain that we should probably be keeping these lists of baskets elsewhere and use the script to reference them in their network location.)

PR #1012, opened 2022-10-10
895 dail decimator restart

IlseFerris:


C-Love (changes requested):
Code looks solid, looks like there is one testing message box. 
Other than that, especially since this is limited use, looks good.

Issue #316, opened 2021-01-05
2022 Standardization Effort - Ensure Worker Signature is in all scripts that CASE/NOTE


IlseFerris:
We need to double check that there is field for worker_signature and that it's mandatory. More information to come in discovery.

C-Love:
We might want to use #125 to guide/support this standardization effort.

MiKaylaH22:
- [ ] ACTIONS	Send SVES
- [ ] ADMIN	UNEA Updater
- [ ] ADMIN	CS Good Cause
- [ ] ADMIN	Individual Appointment Letter
- [ ] ADMIN	Individual NOMI
- [ ] ADMIN	MONT Report
- [ ] ADMIN	On Demand Notes
- [ ] ADMIN	MFIP Sanction FIATer
- [ ] ADMIN	Individual Recertification Notices
- [ ] ADMIN	Review Report
- [ ] BULK	Case Transfer
- [ ] DAIL	Incarceration
- [ ] DAIL	Citizenship Verified
- [ ] DAIL	FMED Deduction
- [ ] DAIL	New Hire
- [ ] DAIL	Postponed Expedited SNAP Verifications
- [ ] DAIL	New Hire NDNH
- [ ] DEU	BULK - DEU-MATCH CLEARED
- [ ] DEU	NOTES - DEU-APPEAL SUMMARY COMPLETED
- [ ] DEU	NOTES - DEU-EBT OUT OF STATE
- [ ] DEU	ACTIONS - DEU-OVERPAYMENT CLAIM ENTERED
- [ ] DEU	ACTIONS - DEU-PARIS MATCH CLEARED
- [ ] DEU	NOTES - DEU-ADH INFO HEARING
- [ ] NOTES	Citizenship Identity Verified
- [ ] NOTES	Disaster Food Replacement
- [ ] NOTES	IMIG - STATUS
- [ ] NOTES	Vendor
- [ ] NOTES	Drug Felon
- [ ] NOTES	GRH NON HRF POSTPAY
- [ ] NOTES	LTC - Hospice Form Received
- [ ] NOTES	LTC - COLA Summary
- [ ] NOTES	Application Check
- [ ] NOTICES	Verifications Still Needed
- [ ] NOTICES	PA Verif Request
- [ ] SHELTER	NOTES - SHELTER-2 PM RETURN
- [ ] SHELTER	NOTES - SHELTER-311
[...]

PR #1017, opened 2022-10-14
#316 BUG due to removal of PF3 between case notes

MiKaylaH22:
Problem: issue reported with scripts case noting in the same case note. 
Resolution: add the PF3 back in between the case notes to ensure a new case note is entered in the following scripts:  

- [x] ACTIONS - DEU-MATCH CLEARED - added back PF3
- [x] ACTIONS - DEU-OVERPAYMENT CLAIM ENTERED - added back PF3 
- [x] NOTES - DEU-ATR RECEIVED - added back PF3
- [x] NOTES - DEU-ADH INFO HEARING - added back PF3
- [x] NOTES - DEU-APPEAL SUMMARY  COMPLETED - updated formatting in case note

### DIFF
--- a/deu/match-cleared.vbs
+++ b/deu/match-cleared.vbs
@@ -850,7 +850,9 @@ IF claim_referral_tracking_dropdown <> "Not Needed" THEN
     IF case_note_only = TRUE THEN CALL write_variable_in_case_note("Maxis case is inactive unable to add or update MISC panel")
     CALL write_variable_in_case_note("-----")
     CALL write_variable_in_case_note(worker_signature)
+	PF3 ' to ensure a new case note is started'
 END IF
+
 start_a_blank_case_note
 IF match_type = "WAGE" THEN CALL write_variable_in_case_note("-----" & IEVS_quarter & " QTR " & IEVS_year & " WAGE MATCH"  & " (" & first_name & ") " & cleared_header & header_note & "-----")
 IF match_type = "BEER" THEN CALL write_variable_in_case_note("-----" & IEVS_year & " NON-WAGE MATCH(" & match_type_letter & ")" & " (" & first_name & ") " & cleared_header & header_note & "-----")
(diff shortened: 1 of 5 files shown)


deu/match-cleared.vbs
```diff
@@ -850,7 +850,9 @@ IF claim_referral_tracking_dropdown <> "Not Needed" THEN
     IF case_note_only = TRUE THEN CALL write_variable_in_case_note("Maxis case is inactive unable to add or update MISC panel")
     CALL write_variable_in_case_note("-----")
     CALL write_variable_in_case_note(worker_signature)
+	PF3 ' to ensure a new case note is started'
 END IF
+
```
IlseFerris, line 855: Should there be a PF3 within the IF...statement starting on line 923?

deu/paris-match-cleared-CC-claim-entered.vbs
```diff
@@ -424,7 +424,9 @@ IF OP_program = "FS" or OP_program_II = "FS" or OP_program_III = "FS" or OP_prog
 	Call write_variable_in_case_note("* Entries for these potential claims must be retained until further notice.")
 	Call write_variable_in_case_note("-----")
 	Call write_variable_in_case_note(worker_signature)
+	PF3 ' to ensure a new case note is started'
 END IF
+
```
IlseFerris, line 429: Same question as overpayment claim entered - Line 465 should there be a PF3

deu/match-cleared.vbs
MiKaylaH22: they should be reviewing it but because it has been there for so long I think it is a good idea to keep it there

IlseFerris (commented):
I have some comments on some of the scripts. I'm not sure these will need to be updated or not.

IlseFerris (approved):
Approved!

Issue #1018, opened 2022-10-14
PND2 Display Limit Bug on NOTES - Application Check


C-Love:
NTOES - Application Check cannot continue if case is in a PND2 basket at the display limit.

PR #1019, opened 2022-10-14
#1018 - NOTES - Application Checkk - bug when PND2 display limit is reached

C-Love:
If case is in a basket where the PND2 display limmit has been reached the script is stuck in a LOOP.


IlseFerris (commented):
Approved!

Issue #1039, opened 2022-11-03
DAIL - CSES - Bug when no approved version of SNAP


C-Love:
Bug report on a case that does not have an approved version of SNAP. Need to update functionality to support unapproved SNAP.

C-Love:
This is done and now very well tested.

PR #1040, opened 2022-11-03
#1039 - DAIL - CSES Scrubber Bug on a REIN case no approved budget

C-Love:


IlseFerris (approved):
Did not test, but it looks solid. I approve thee!

Issue #89, opened 2020-07-01
NOTES-IMIG-SPONSOR INCOME update to handle for multiple sponsors 


MiKaylaH22:
Didn’t  we have a second script created which allowed for income of more than one   sponsor to be counted and deemed in the event there is more than one sponsor   ?     I am preparing some information for Hot Topics in response to a request   resulting from a QC error recently.
--Louise

MiKaylaH22:
RE: process
 When a SAVE report is run at application, review or change in immigration status, we need to determine if the non-citizen is in a sponsored category (please see the Non-Citizen Cheat Sheet on SharePoint).  If so, the name of the sponsor(s) must be verified via the same SAVE report as well as their “Grant or Entry Date”.  

Once you have that information you need to check if there is a way of removing the sponsorship requirements such as the person being granted naturalized citizenship (they’ll report this usually); if old enough, run TPQY to verify if they have earned 40 SSA work quarters; or if the sponsor passed away.  Other than the TPQY, the rest can be gleemed off case notes or during the interview.

If that fails we need you to see if they are possibly eligible for an Indigent Exemption, review the reporting form information with them ensuring they understand that their name, case info and the sponsor(s) info will be reported to MN DHS who in turn can share that with USCIS.  Normally, if they wish to proceed we’d need a signature on the form for cash assistance along with a written statement as to why the sponsor isn’t supporting them.  They must also be considered precariously housed (at risk of losing housing such as eviction, etc.) though they must be able to demonstrate how they will pay future rent obligations.  Due to the current situation our state finds itself in, we are temporarily waving the required signature and possibly the housing requirement.  Send an email with the information (or the date of your note documenting it) to Rene.Rodriguez@hennepin.us or Jillian.Kyles@hennepin.us. 

Once we have the information, we will email you and add a note to your case showing program approved, date of the exemption, etc.  The form you attached to your email is the reporting form we need you to explain to her.  Verbally get the approval as well as the statement regarding the sponsor’s failure to support.  Should they refuse to accept an indigent exemption, you will need to verify or request the sponsor’s and the sponsor’s spouse if married, income/assets.  Hope this doesn’t make things more confusing!  


Thanks for contacting HP.Immigration!  

René Rodríguez

PR #1047, opened 2022-11-14
89 notes imig spon

IlseFerris:
Both issue #89 and #1045 are in this pull request. 
Will update COMPLETE LIST OF SCRIPTS directly to change the NOTES - IMIG - STATUS text from all caps.

### DIFF
--- a/notes/imig-sponsor-income.vbs
+++ b/notes/imig-sponsor-income.vbs
@@ -170,6 +170,13 @@ Do
         exit do
     Else
         PF3 'save case note
+        'blanking out variables for the next unique sponsor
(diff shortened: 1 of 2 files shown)


C-Love (changes requested):
I think we should blank out the income and count fields if we loop. (primary_sponsor_earned_income, spousal_sponsor_earned_income, primary_sponsor_unearned_income, spousal_sponsor_unearned_income, spousal_sponsor_unearned_income) Not sure about 'number_of_sponsored_immigrants' as that could be the same on each loop BUT it might not be.

C-Love (approved):
Looks good.

PR #1049, opened 2022-11-14
1014 notes ltc 5181

IlseFerris:


C-Love (changes requested):
There is no handling for if 3 different names are entered. To clarify what should be on the panel.

Also, this is optional but these dialogs are U-G-L-Y

C-Love (approved):
Looks good.

Issue #622, opened 2021-10-05
Project Time Bandit - BZST Time Tracking via scripts


C-Love:
Lets make our time tracking an annual thing and updated/viewed via scripts!

C-Love:
Add  + and - button to switch dates in Hours Report..

C-Love:
Adding discovery on  functionality to support time off tracking.

MiKaylaH22:
my requests for time off are not being received by our HSS. I am getting the success messages and have confirmed I am on the master branch

C-Love:
This functionality isn't done - iI am not surprised it isn't working fully. We will need to set up a time for functionality review and testing.

C-Love:
New plan - turn off the access to the start of the request time off functionality and complete/close this issue.

C-Love:
Add functionality to futurize the date creations within the script so that we don't have to hard code change for dates.

C-Love:
This is DONE!

PR #1050, opened 2022-11-14
#622 BZST Time Tracking Initial Functionality Complete

C-Love:
Updates in this pull:

- Commented out the 'Leave Request' functionality. We are dropping that for now but might resume it in the future.
- Moved to having one Time Tracking Per Person - no longer by year
- Dynamic creation of the list of weeks, pay periods, and months
- Using an 'Active Time Period' for lookback for speed and ease of use. 
- Automatic archiving by year done by the script.


IlseFerris (approved):
I approve thee!

Issue #992, opened 2022-09-29
NOTES - Emergency - EGA Screening - Make screening MsgBox a Dialog


C-Love:
Update the functionality of NOTES - Emergency for the EGA Screening to appear as a Dialog instead of a MsgBox so that the buttons can be custom wording to clarify what will happen next in the script.

This will help users to correctly use this functionality to best support their work.

C-Love:
This appears to be functioning.

PR #1052, opened 2022-11-14
#992 - NOTES - Emergency - Move screening display to Dialog

C-Love:


IlseFerris (changes requested):
Just needs a changelog, then approved.

Issue #983, opened 2022-09-15
DEU - BULK - MATCH CLEARED - enhanced handling for singled user 


MiKaylaH22:
remove line 213 - excel_row = excel_row + 1
remove line 237 - excel_row = excel_row + 1
add  excel_row = excel_row + 1 under END IF 238 
add IF entry_record = 0 then script_end_procedure("No information found on excel sheet")

IlseFerris:
Enhancement required in the script to provide more information when cases have already been cleared or matches have not been found.

PR #1056, opened 2022-11-15
#983 DEU - BULK MATCH CLEARED: Fixed bug with incrementing excel row

IlseFerris:
Added changelog
Removed worker_county_code
Removed MH as authorized script user
Updated removed 2 if statements and made 1 if statement for adding information to an array. Removed excess excel_row counts that was causing cases to be skipped. Added entry record 0 script end per issue request. Updated msgbox to user at excel output to have more information. Reviewed closing issue documentation


C-Love (approved):
This is fine

Issue #980, opened 2022-09-15
New Script - Display of Issued Benefits - UTILITY for Resident Conversation


C-Love:
This would be a display of benefits issued in a dialog to increase communication to residents about their benefits.

Use MONY/INQX functionality from PA Verif.
Read out active/pending programs. 
Maybe read the ELIG results for CM +1 from active programs to see the ongoing benefit detail in the dialog. 

This is a tool for supporting communication with residents.

C-Love:
Script is released closing.

PR #1061, opened 2022-11-16
#980 New Script and Functionality - Display Benefits

C-Love:
You will not actually be able to run the new script unless you add a 'run_locally = true' to it since the functionality is in the FuncLib and we don't have the CLOS updated.


IlseFerris (approved):
I approve thee!

Issue #1060, opened 2022-11-16
REPT/REVS: Add better date handling when user isn't selecting CM+2 


IlseFerris:
The date handling is not intuitive, and relies on the worker having the correct footer month/year already set up. We can do better than this.

IlseFerris:
End user reported success using the script/updates. Closing issue.

PR #1075, opened 2022-11-26
#1060 rept revs

IlseFerris:


C-Love (approved):
This is way smoother code. Nice.

Issue #1042, opened 2022-11-09
NOTES - Application Check - Update for Day 30 Assignments - Direction for Denials


C-Love:
Now that Day 30 work is assigned to all through ES Workflow, we are having some needs to help support the correct processing of these cases. 

The general process steps will be:

- Find if at Day 30
- Ask if an interview has been done (script check PROG first)
- If 'No' Interview: check for appt notc and nomi dates, advise of closure, NO NOTE if closure is possible
- If 'Yes' Interview: ask about verifs and verif requests. Advise of process or closure. NO NOTE if closure is possible

This should help to support worker action consistency. 

This is related to work on #803

C-Love:
Need to add handling for pending 60 days for MSA/GRH if DISA and interview is done.

C-Love:
No errors reported,, appears to be working.

PR #1083, opened 2022-11-30
#1042 NOTES - Application Check - Allow for 60 Days for MSA/GRH

C-Love:
These updates will allow the script to run the standard application check for MSA/GRH cases that have an interview between days 30-59. 

The logic gets a little complicated here but I have a list of some cases for testing if you want them.


IlseFerris (approved):
I approve thee. Looks good. Didn't run.

Issue #1067, opened 2022-11-21
DAIL Scrubber - create functionality to get the CORRECT DAIL to the Top of the list


C-Love:
We have had some issues where the DAIL Scrubber navigates around and the wrong message ends up being used for additional navigation or actions/reading.

Idea: Read the WHOLE line and every time we move from the DAIL, add a function to do loop through all the lines and find the right one to bring to the top. This would be a function that would be used every time we navigate to and away from the DAIL.

Issue #476, opened 2021-05-26
DAIL- SCRUBBER - navigation to ensure we are in production


MiKaylaH22:
Add navigate DAIL/WRIT to ensure worker is not in INQUIRY
- [ ] CATCH ALL
- [ ] CITIZEN VERIF
- [ ] COLA REVIEW & APP
- [ ] CSES SCRUBBER
- [ ] DISA MSG
- [ ] ES REF MISSING
- [ ] FIN ORIENTATION MISSING
- [ ] FMED DEDUCTION
- [ ] INCACERATION
- [ ] LTC REMEDIAL CARE
- [ ] MEDI CHECK
- [ ] NDNH
- [ ] NEW HIRE
- [ ] OVERDUE BABY
- [ ] PAPERERLESS DAIL
- [ ] POSTPONED EXP

IlseFerris:
Idea - update DAIL SCRUBBER with in_inquiry = true/false, then update the DAIL SCRUBBER to continue or script end depending on the process.

C-Love:
Functionality in place and testing well. Next step is to assess each DAIL Scrubber individual script to script_end_procedure where appropriate. Good for little issues day.

C-Love:
this appears to be working without being noticed.

PR #1088, opened 2022-12-09
#1067 & #476 DAIL Scrubber Navigation

C-Love:
Updates to the DAIL scrubber that will:

 - Allow the script to find the correct message better - this is a new function that can be added to any DAIL scrubber script.
 - Determine if the script is being run in inquiry. Each functionality can be evaluated to determine if the script should stop when run in INQUIRY.


IlseFerris (approved):
Looks good. I approve thee!

PR #1125, opened 2023-01-12
#476 DAIL Scrubber - script end if in inquiry sometimes

C-Love:
Reviewed the DAIL Scrubber individual functionality and added a script end procedure for the scripts that require the ability to take action.

The script end procedures are all in the main DIAL Scrubber.


IlseFerris (approved):
I approve thee!

PR #1126, opened 2023-01-12
#486 Closing Issue Documentation: added reviewing dialog

IlseFerris:
Also added a date to see if we have the most up-to-date version when reviewing code, and updated the MAXIS entry line to fit in line with all other descriptions (it was getting to be long).


C-Love (approved):
Yep

Issue #1094, opened 2022-12-22
DAIL - New Hire NDNH Handling for Closed Cases


C-Love:
KW is clearing CCL DAILs, and the NDNH New Hires require INFC work that the script is helpful for automating. However, the details in the note and the functionality of the script are designed to support active cases. KW has requested an update to NDNH Dail Scrubber to support these specific process requirements. 

Idea:
 - Have the script read if we are working from CCL
 - If in CCL, have some separate functionality and a different dialog
 - Update the CASE/NOTE to be clear that we are clearing the DAIL with limited case review and no update, including no ECF review.

Information that is important to gather/NOTE:
 - Amount of time the case has been closed (we should divert the functionality if less than 4 months).
 - Basic reason for closure
 - If the job was known or not
 - INFC resolution information
 - JOB detail from the DAIL

KW has a large list to clear and the current script does not provide a clear CASE/NOTE. A temporary workaround was requested, to just remove the piece about ECF, but this still creates a less efficient script run and a NOTE that is not distinguishable from a standard New Hire Resolution note. Recommend completing this quickly so that KW can complete her cleanup in Jan.

C-Love:
This is released and for a single user. Done.

PR #1127, opened 2023-01-12
#1094 - Closed Case New Hire DAIL Support

C-Love:
Built specifically for KW to complete cleaning of CCL DAILs.


IlseFerris (changes requested):
Just update the changelog and then we're good to go!

Issue #1105, opened 2023-01-06
NOTES - Application Check - add end script message about PND2 Display limit


C-Love:
If a case cannot be found because the case is not visible on PND2 due to a display limit being reached, the end message should be specific to the fact that this case is in an X-Number that is over the display limit and there may be some impact on script operation.

This end message verbiage will need to have some logic to identify if the display limit was hit and if the case cannot be found.

C-Love:
No errors reported, appears to be working.

PR #1129, opened 2023-01-12
#1105 NOTES - Application Check - messaging for PND2 Display Limit

C-Love:
Added specific messaging around the fact that sometimes there is a Display Limit issue that causes us to not be able to read Pending details. The script end procedure is not clear for these cases because it states they are not pending, which may not be true. 

Additional clarity added to the end message here.


IlseFerris (approved):
I approve thee!

Issue #1090, opened 2022-12-16
ACTIONS - Job Change Reported BUG when start date is prior to the application month


C-Love:
Reported that the script would not accept a March start date of employment for a case that was inactive at the time the job started. 

Details on case SNAP denied 09/19,, Cash, EMER and SNAP denied 07/22. Pended in 12/22 for SNAP. Job started in 03/22.

Discovery and resolution needed.

C-Love:
Appears to be fixed.

PR #1133, opened 2023-01-12
#1090 ACTIONS  Job Change Reported new job before first month

C-Love:
If we have a job reported that started before the APPL month, everything breaks. Handling to resolve that.


IlseFerris (approved):
I approve thee!

Issue #260, opened 2020-11-04
PEPR assignment for BULK DAILS when we hit the STATS we hit an error msg


MiKaylaH22:
Happens more often then not line 1 error

C-Love:
I have run this script in 6 different worker number combinations and checkbox combinations. Cannot repeat this error. Update to match the project convention of iterating the excel row. No additional changes identified.

C-Love:
Could not recreate the error. Added some stability with matching the incrementing convention. No 'fix' in place., could not discover.

PR #1134, opened 2023-01-12
#260 BULK - DAIL Report - update excel row iteration

C-Love:
There have been reports of the stats section erroring when being created. Unable to reliably reproduce this error. The convention for iterating the excel row doesn't match the project convention and this may cause a stability issue, updating to prevent this possibility.


IlseFerris (approved):
I approve thee!

Issue #1097, opened 2022-12-22
NOTES - CAF - Expedited Determination - BUG on adding UNEA to the Exp Income Calculation


C-Love:
The NOTES - CAF script should be counting all of the UNEA when making a determination for Expedited, but it was only counting the MAIN UNEA sources. 

Updated the initial count functionality to include ANY UNEA.

C-Love:
This appears to be working correctly now.

C-Love:
This is broken in a new way now

C-Love:
This is done!

PR #1135, opened 2023-01-12
#1097 NOTES - CAF - determining the starting UNEA amount for EXP

C-Love:
Created a new variable to collect all of the UNEA income and add it to the rest of the income.


IlseFerris (approved):
I approve thee!

Issue #1079, opened 2022-11-26
Update function: Create_array_of_all_active_x_numbers_by_supervisor


IlseFerris:
BULK - REPT-REVS has a local function of this function that has an exit do condition that is not in the FuncLib version: 
```
EMReadScreen end_check, 9, 24,14
If end_check = "LAST PAGE" Then Exit Do
```
To do to resolve issue: 
- [x] Replace function in Funclib
- [x] Remove local versions of function

IlseFerris:
No issues reported. Closing issue.

Issue #1078, opened 2022-11-26
FuncLib: Update PF functions that shift with lables/comments


IlseFerris:
For ease of use in coding we should update the PF functions in the FuncLib to say which function they are if there is a shift needed in BlueZone (example: PF20 is Shift + PF8)

Issue #1077, opened 2022-11-26
Remove function FIX_READ_DATA


IlseFerris:
We are not using this in our project.

Issue #1076, opened 2022-11-26
Update FuncLib Call


IlseFerris:
The current FuncLib call has text from the initial DHS project, and the Desert Island Version is also not online. Add: 
- [x] General FuncLib file with BZST contact info
- [x] Desert Island FuncLib file

PR #1137, opened 2023-01-13
All the func lib

IlseFerris:
Cleaning up the functions and function calls. 
#1079 updated the function in the FuncLib, and removed local calls,
#1078 updated instructions for PF13-24
#1077 removed/retired
#1076 updated and added DESERT ISLAND FuncLib call


C-Love (approved):
Looks good.

PR #1147, opened 2023-01-26
1139 ecf verbiage updates

IlseFerris:
All initially tested.


C-Love (approved):
I did not run these but since they are pretty much all verbiage change, they all look good.

Issue #1104, opened 2023-01-04
Case Assignment Subsequent Application Functionality


C-Love:
We need to either adjust NOTES - Application Received to include support for cases with a subsequent application OR we need to create an additional script to support subsequent applications. The second option would also likely include renaming NOTES - Application Received.

Basically we need to handle for different processes that might occur when an application is received in the agency.

PR #1148, opened 2023-01-26
#1104 New Script - NOTES - Subsequent Applications

C-Love:
New script file to support case assignment work when a subsequent application is received.

I plan to pull this in Monday morning for testing by CA.

This pull does not have updates to CLOS or the CA menu updates. I will complete those when I am ready to pull the changes in.

### DIFF
--- a/notes/application-received.vbs
+++ b/notes/application-received.vbs
@@ -225,6 +225,64 @@ IF IsDate(application_date) = False THEN                   'If we could NOT find
     application_date = pnd2_appl_date
 End if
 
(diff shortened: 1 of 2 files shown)


IlseFerris (approved):
I didn't run this, but since it was tested together on LID it's totally fine. I didn't see anything in the code that was sus. 
Will we need to update the desert island to have the stand-alone subsequent application support available, or will CA just be directed to use APP REC'D in that situation? 
Long story short, I approve thee!

Issue #209, opened 2020-09-28
Create new BULK Report to identify active HC cases that are missing an approved budget for the current month


C-Love:
There are a number of cases in MAXIS - likely a very LARGE number of cases that have no current budget in MAXIS. The idea of this script would be to review the cases and find all of the cases with no current budget. The script will then create a budget that matches the currently eiligibility for HC.

This could also align the budgets with review dates.

This way EVERY HC Case will have a budget! We can run this monthly to keep up with it. The first month will be the worst.

Some additional thoughts/notes:
![image](https://user-images.githubusercontent.com/13418322/94445449-1167a400-016d-11eb-8fb3-9ca8acea4dcb.png)

C-Love:
This is no longer a repair process, just identification.

C-Love:
We need to create this BULK script to have available and then make awareness.

C-Love:
This script is done and ready.

PR #1150, opened 2023-01-26
#209 BULK - Health Care Budget Report - New Script

C-Love:
New script to create a report of HC cases that require review as they do not have an approved budget in a specific month.


IlseFerris (approved):
I approve thee!

PR #1177, opened 2023-02-27
1169 notes mets retro hc

IlseFerris:


C-Love (approved):
The only thing to possibly add is information to the end message that a document has been opened to copy and paste into METS.

Issue #1161, opened 2023-02-10
NOTES - OVERPAYMENT: Create collectable/non collectable indicators


IlseFerris:
Request from tester AR: 

> I just completed a non-collectible claim (overpayment) for a case.
> There is no option on the Overpayment case note to select collectible or non-collectible.
> This may be a good enhancement as it needs to be very clear for the next worker so they don’t accidentally tell the resident they have an overpayment.
> I had to put it in manually on the case note.
> The resident does not receive an overpayment notice for non-collectible claims, only for collectible claims.

This is a great idea. Will add collectable/non collectable indicator to the claims.

C-Love:
This appears to be working just fine.

PR #1189, opened 2023-03-01
#1161 NOTES - Overpayment - include collectible selection

C-Love:
To ensure that the information about collectible fits on the same line of the CASE/NOTE I had to change the note information from 'through' to 'thru'. I know this is more informal but the meaning is clear and it makes everything fit.


IlseFerris (approved):
I approve thee!

Issue #51, opened 2020-06-10
Update to NOTES - SMRT to match new State Requirements


C-Love:
The state has updated the requirements for a SMRT referral. We will update the script NOTES - SMRT to handle for these updates. 

1. MA must be active or pending to submit a SMRT referral. Add a dropdown to both the 'Initial request' and 'ISDS Referral Completed' dialogs to indicate if this person is active or pending MA.
2. Must check SMI for disability determination FIRST. Add droplist that worker needs to indicate SMI has been reviewed.
3. Get an image of the ECF referral form and the ISDS Referral so we can be sure all the correct items are added. Mostly looking at 'Programs Requested'.
4. If this is for a waiver the MN Choices assessment completion date or scheduled date will need to be entered/captured.
5. We may need a better way to track these to ensure the pending MA cases aren't denied and referrals aren't refused. Sent Email to Shannon Stockdale about counts.

C-Love:
Add Redirect to the MC Menu to access the NOTES - SMRT script.

Add changelog to NOTES - SMRT to request report for any updates desired.

Add in script error reporting to NOTES - SMRT to get more ideas and enhancements for this script.

MiKaylaH22:
Question was asked by KNC - MA must be active or pending to submit a SMRT referral - where did this information come from?

MiKaylaH22:
This information came from HC policy Meeting 2019 and the parts that are documented are the parts that we will support.

MiKaylaH22:
- [ ] Must check SMI for disability determination FIRST.  
- [ ] Ask the question of is client pending or active on MA? since it is not documented but we know it is needed. 
- [ ] Add droplist that worker needs to indicate SMI has been reviewed.
- [ ] Get an image of the ECF referral form and the ISDS Referral so we can be sure all the correct items are added. Mostly looking at 'Programs Requested'.
- [ ] If this is for a waiver the MN Choices assessment completion date or scheduled date will need to be entered/captured.
- [ ] We may need a better way to track these to ensure the pending MA cases aren't denied and referrals aren't refused. 
- [ ] If the reason is not listed, call SMRT at 800-235-7396 or 651-431-2493 to get approval. 
REMEMBER  to Check the box next to SMRT Approved and add a comment before submitting the referral.

MiKaylaH22:
Would it be possible to do the following to the NOTES > SMRT > Determination Received:
[...]

PR #1196, opened 2023-03-07
#51 NOTES - SMRT workflow process update

C-Love:


IlseFerris (approved):
I tested this and it looks/works great. Just for awareness I had an exception error the 1st time I ran it but it didn't happen again, so I'm thinking it's just  fluke. 
I approve thee!

Issue #560, opened 2021-08-16
NOTES - Interview - Add handling different Application Forms


C-Love:
Currently the Interview is built for ONLY one version of the CAF form (which includes MN Benefits and ApplyMN) in #62. Need to add functionality to support the questions asked and discussions needed for these forms: 

- [ ] Sr SNAP Application
- [ ] Combined AR
- [ ] Other versions of the CAF by release date

C-Love:
This appears to be functioning well, no errors reported

PR #1200, opened 2023-03-09
#560 NOTES - Interview - removal of redirect or script end for NOT CAF

C-Love:
So apparently we were redirecting to Interview Completed... ugh

Changed it to be just a message and continue on to work.


IlseFerris (approved):
I approve thee!

PR #1205, opened 2023-03-13
1180 syntax error in check for password

IlseFerris:
Only updates on the do...loop coding. Did not add changelogs since it's background functionality.


C-Love (approved):
Looks good. I did not run each of these but it looks solid.

PR #1206, opened 2023-03-14
1187 stop script house keeping

IlseFerris:
No changelog since this is behind the scene updates. Limited updates to only replacing StopScript. All tested for initial errors.


C-Love (approved):
Looks good to me

Issue #311, opened 2020-12-29
AVS Recertification Updates


IlseFerris:
Associated with issue #140. 

Currently the AVS script for eligibility workers supports only Application and Change-in-Basis situations per the current policy. When the policy changes (date unknown at this point), the script will also need to support Recertification processes as well. 

Updates will include: 
- [x] NOTES - AVS: HC Process drop lists update to include Change in Basis
- [x] NOTES - AVS: Remove default for APPLICATION in HC Process
- [x] NOTES - AVS: Remove script end procedure for HC_process = "Renewal"
- [x] UTILITIES - QI AVS REQUEST: Add option for Renewal/Change in Basis when requesting submission.
- [x] #1209
- [x] UTILITIES - QI AVS REQUEST: Add msgbox for training cases vs. sending email.

Not including initial task from issue of: Verbiage for WCOMS: "[Enrollee’s name] coverage will end because you did not return the completed Authorization to Obtain Financial Information from the Account Validation Service (AVS) form (DHS-7823). If you return the completed form by the last day of the month of this notice, your coverage may be reinstated."
Cannot find the procedure, and may have been removed due to the PHE.

PR #1210, opened 2023-03-20
#311 and #1209 Issues

IlseFerris:
Enjoy :)

### DIFF
--- a/utilities/qi-avs-request.vbs
+++ b/utilities/qi-avs-request.vbs
@@ -43,6 +43,7 @@ changelog = array()
 
 'INSERT ACTUAL CHANGES HERE, WITH PARAMETERS DATE, DESCRIPTION, AND SCRIPTWRITER. **ENSURE THE MOST RECENT CHANGE GOES ON TOP!!**
 'Example: call changelog_update("01/01/2000", "The script has been updated to fix a typo on the initial dialog.", "Jane Public, Oak County")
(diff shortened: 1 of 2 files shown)


utilities/qi-avs-request.vbs
```diff
@@ -161,50 +152,37 @@ const client_DOB_const   		    = 5 '=  Date of Birth MEMB
...
+
     EMReadScreen client_DOB, 10, 8, 42
+
     EMReadscreen client_SSN, 11, 7, 42
     If client_ssn = "___ __ ____" then client_ssn = ""
-    last_name = trim(replace(last_name, "_", "")) & " "
-    first_name = trim(replace(first_name, "_", "")) & " "
-    mid_initial = replace(mid_initial, "_", "")
+
     EMReadScreen client_age, 2, 8, 76
```
C-Love, line 176: This can sometimes be over 99, and we have to read 3 spaces, otherwise the math will be weird later.

C-Love: Nevermind - we aren't doing anything with the age here - maybe delete this.

C-Love (changes requested):
So, that was fun. There are a couple of notes here to review. Nothing major, it could probably go ahead as is if we wanted to.

C-Love (approved):
So fresh and so clean... clean

Issue #129, opened 2020-07-26
Combine all Health Care Apps


C-Love:
These should all really be one script.

![image](https://user-images.githubusercontent.com/13418322/88483697-410cec80-cf2f-11ea-9745-7a91162e0548.png)

IMIG-EMA also

This is on hold until some things calm down.
Flagging for Q1 of 2021 but we can easily move it.

C-Love:
Account for MA EPD

C-Love:
Make sure to add Navigate to MMIS button or other MMIS functionality - work this out.

C-Love:
Other scripts to incorporate:

- NOTES - IMIG - EMA
- NOTES - METS RETRO HEALTH CARE
- NOTES -  Health Care Transition (possible - needs research)

C-Love:
Use DASH User Roles to help with some of the selection defaults and options

IlseFerris:
Feedback from user: "I would like a field for the EMMA panel". In future updates reading from EMMA functionality will be reviewed.

C-Love:
Better handling for retro requests on HCAPP processing. 
 - A notes line for each specific Retro Month - one for each - up to three.
 - Better details around verifications and if they are for the retro request or if the retro request has no verifications needed.
Request from L.Messer - contact for more information/feedback/details

C-Love:
4/13/22 PIX meeting had a presentation about differend MHCP Applications - https://www.dhssir.cty.dhs.state.mn.us/MMIS/HCEPIXmeeting/2022.04.13%20PIX%20PPT.pdf (PowerPoint)
http://stream5.video.state.mn.us:8080/MNIT/_hcepix_04-13-22.html (Video)

C-Love:
The current PHE is set to expire after July 14 2022. If this is not extended HC renewals COULD resume for October 2022. This could be later and if the PHE is extended, it would likely be extended to Mid-October which would mean the renewals would be months after that time. It appears the earliest renewals could happen would be October. Recommend to move this to Q3 as we should know by that time if this work would be needed for cases to be processed in Sept for Oct renewals.

C-Love:
There is a good slide about MA-LTC Application forms in the 05/2022 PIX - Link to the Powerpoint in SIR:https://www.dhssir.cty.dhs.state.mn.us/MMIS/HCEPIXmeeting/2022.05.11%20PIX%20PPT.pdf

C-Love:
[...]

PR #1223, opened 2023-04-01
1191 desert island elig summary updates

IlseFerris:
This will need to sit for a bit until Health Care #129 are ready to be pulled in. Otherwise the HC script won't work.


C-Love:
Merge conflict - committed changes directly to MASTER

Issue #1218, opened 2023-03-23
Restructure script for HC Applications


C-Love:
This script is a part of issue #129. 

This script will replace HCAPP and LTC - Application Received to document review of and application for HC. 

This new functionality will support functionality in #849 to document a case is at 'STANDARD POLICY'.

Additional notes to be added as plan is in place.

C-Love:
Retire 
NOTES - HCAPP
NOTES - IMIG EMA
NOTES - Application Received

Previously retired that will remain retired 
NOTES - HC Renewal
NOTES - LTC Renewal

C-Love:
This new script is available and appears to be working.

PR #1236, opened 2023-04-11
#1218 - Releasing HC Evaluation to Testing - redirects in 3 scripts

C-Love:
Select testing file redirect to the new HC Evaluation script from:

- HCAPP
- LTC Application Received
- IMIG EMA 

These scripts will all retire in the next weeks and we need testing.


IlseFerris (approved):
I approve thee!

PR #1245, opened 2023-04-16
1240 hc screning application check

IlseFerris:


C-Love (approved):
This looks good. Get this screening out of here!

I did not actually run it but the code is good.

PR #1274, opened 2023-04-25
1225 update redirect files

MeganGeissler:


IlseFerris (approved):
Looks great. I connected with @C-Love and she confirmed that NAV scripts were out of scope. 
I approve thee!

Issue #1275, opened 2023-04-27
NOTICES - PA Verif Request - remove person specific direction for Public Housing Form and direct users to HSR Manual


C-Love:
The HSR manual has been updated to direct workers to complete PA Verification Forms themselves instead of sending them to individual workers. We are going to align this script.

Additionally, we might need to update the selection verbiage to make it not specific to Public Housing.

IlseFerris:
Knowledge Coordinator OK'd the change. Closing issue.

PR #1289, opened 2023-05-02
#1275 Notices - PA Verif Request

MDR22:
Updated the dropdown option for "PHA (Public Housing form)" to "Request from Third Party (including Public Housing Agencies)" to align with HSR manual. Updated the instructional MsgBox to match the HSR manual instructions.


IlseFerris (approved):
Tested the new option without issue, and text looks good. I approve thee!

Issue #452, opened 2021-04-27
Hot Topic Power Pad Button to Open a list of BZST Hot Topics


C-Love:
We publish information about our scripts in Hot Topics and have space in the CLOS to save the dates and URLs for Hot Topics by script.

This script project would:
1. Create a dialog that uses the CLOS to have a dynamic display of all of the scripts information and Buttons to open a link to each article. 
2. Have the dialog still have a link to the main Hot Topic Page

We will also need to review all the Hot Topic articles for BZST and update CLOS to put in our Hot Topic article history.

IlseFerris:
Update: 

- [ ] Phase One: Link to the BZ SharePoint Page, change button name to reference BZ Hot Topics
- [ ] Phase Two: 
    - Create a dialog that uses the CLOS to have a dynamic display of all of the scripts information and Buttons to open a link to each article.
    - Focus on ONLY BlueZone Script articles.

IlseFerris:
Phase 1 is complete.

IlseFerris:
Added announcement, closing issue.

PR #1293, opened 2023-05-04
#452 BlueZone Hot Topic Display

C-Love:
New script that will display the Hot Topics we create about the BlueZone Scripts in a dialog. this will allow users to connect directly to the scripts, running them and opening the instructions.


IlseFerris (approved):
I love this so much! I very much approve thee! 

One comment though - The instructions and Favorite text is a bit wonky. I'm thinking this is due to space, but maybe we can think out an option for this later.
image

C-Love:
@IlseFerris how about this?
image

PR #1304, opened 2023-05-09
#452 UTILITIES - Hot Topics - add to CLOS and some formatting updates

C-Love:
Added the Hot Topics script to the CLOS so that we can run it from the UTILITY Menu as well. 

Additional bugs found in the formatting of the display;
- dialog size calculation
- Capitalization in the URLs for the Hot Topics

Bug with a constant name being redefined, changed the name to be HT specific.


IlseFerris (approved):
I approve thee!

Issue #1280, opened 2023-04-28
Update the Functionality of find_user_name to not break everything


C-Love:
We have finally learned the actual issue when workers cannot use the CAF or Interview scripts.

It is an Outlook issue when trying to pull the user name.

We are going to test using the staff table to pull this information as a first option and have the script read this detail from somewhere other than Outlook. Once testers confirm they have access to read this table, we can update the function for all.

C-Love:
It does not appear that users have access to the table with user data. 

We need to find a new solution. Possibilities:
 - See about additional access to select for this table by all users
 - Create a process to download this information to a txt file in the T Drive that is accessible to all so the names can be searched from there.
 - Create an option when reading from outlook to see if it is open or not.

We are going to do discovery on access to the table first. We will leave the test script in place until discovery on access is completed or abandoned.

C-Love:
Additional hard coded override from the Table details for AH QI member to support name chnage.

C-Love:
This done - we will follow up on some data needs in #1361

PR #1324, opened 2023-05-19
#1280 Change find_user_name function to not use Outlook

C-Love:
Pulled functionality from a testing script that we used to ensure user access that will read user name from a table instead of from Outlook.

This will resolve the Line 1 error we see on the CAF and Interview scripts.


IlseFerris (approved):
Not tested, but looks good. I approve thee!

Issue #1254, opened 2023-04-24
ACTIONS - TRANSFER CASE: Ensure dates will work for specific formats


IlseFerris:
Periodic issues have been reported with out of county transfers erroring out. One issue that was isolated was issues where staff enter a date that is a MM/YYYY vs. having the month, day and year. Need to add handling with DateFormat to handle for those dates that are entered. Already checking for isDate, but needs more handling.

IlseFerris:
No issues reported, Closing issue.

PR #1325, opened 2023-05-23
#1254 ACTIONS - TRANSFER CASE: Ensure dates will work for specific formats

MDR22:
Updated date validation to ensure dates are in MM/DD/YYYY format. Fixed typos within error messages. Updated tab order on dialogs.


actions/transfer-case.vbs
```diff
@@ -199,7 +199,7 @@ Do
     Do
         'Inside the do...loop for the calculate_button
         Dialog1 = ""
-        BeginDialog Dialog1, 0, 0, 346, 260, "Out-of-County Case Transfer for #" & MAXIS_case_number
+        BeginDialog Dialog1, 0, 0, 346, 260, "Out-of-County Case Transfer for # & MAXIS_case_number"
```
IlseFerris, line 202: Line 202 - The string will need to be adjusted to exclude the MAXIS_case_number. See the diff for the example.

MDR22: Good catch - reverted back to original.

PR #1328, opened 2023-05-30
#129 NOTES - HC Eval - add recert option

C-Love:
Updates to allow the script work for recertifications.


MDR22 (approved):
Changes look good to me!

Issue #1333, opened 2023-05-31
Notes - Health Care Evaluation - Add MISC - VERIFY EX PARTE script


MDR22:
Incorporate MISC - VERIFY EX PARTE script into HC Eval script

C-Love:
This is completed and the functionality is available in HC Eval.

PR #1334, opened 2023-05-31
#1333 notes - health care evaluation - add misc - verify ex parte script

MDR22:
Incorporated MISC - Verify Ex Parte script into HC Eval script.

### DIFF
--- a/notes/health-care-evaluation.vbs
+++ b/notes/health-care-evaluation.vbs
@@ -2817,6 +2817,7 @@ form_selection_options = form_selection_options+chr(9)+"Breast and Cervical Canc
 form_selection_options = form_selection_options+chr(9)+"Minnesota Family Planning Program Application (DHS-4740)"
 form_selection_options = form_selection_options+chr(9)+"SAGE Enrollment Form (MA/BC PE Only)"
 form_selection_options = form_selection_options+chr(9)+"Screen Our Circle Form (MA/BC PE Only)"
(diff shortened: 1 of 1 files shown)


notes/health-care-evaluation.vbs
```diff
@@ -2975,13 +2978,610 @@ DO
...
+            If HC_form_name = "No Form - Ex Parte Determination" AND ltc_waiver_request_yn <> "N/A - Ex Parte Process" THEN err_msg = err_msg & vbCr & "* Select 'N/A - Ex Parte Process' for the LTC/Waiver Services field."
 			If trim(worker_signature) = "" Then err_msg = err_msg & vbCr & "* Enter your name to sign your CASE/NOTE."
 			IF err_msg <> "" THEN MsgBox "*** NOTICE!!! ***" & vbNewLine & err_msg & vbNewLine
 		End If
 	LOOP UNTIL err_msg = ""
 	CALL check_for_password_without_transmit(are_we_passworded_out)
 Loop until are_we_passworded_out = false
 
+'Add ex parte SQL connections and script
+If HC_form_name = "No Form - Ex Parte Determination" Then
```
C-Love, line 2991: All of the functionality needs to be nested in this if statement so that this and only this run when Ex Parte is selected.

MDR22: Makes sense - I made that change

Issue #1331, opened 2023-05-31
NOTES - Applications Received bug when REPT/PND2 is too big - cannot find app date


C-Love:
Report of a bug on Application Received if PND2 is too big and the case is not displayed on the transmit past the warning. This throws an error because we try to do a date thing with a variable that is never pulled if we cannot get to PND2 with the case displayed. 

The resolution to this is to have the script end when this situation is reached.

C-Love:
This is working! Yay

PR #1342, opened 2023-06-06
#1331 NOTES - App Recvd - end script if case not found on PND2

C-Love:


IlseFerris (approved):
Looks good but I didn't run to test. I approve thee!

Issue #1343, opened 2023-06-06
Add new ABAWD DAIL Message to DAIL Scrubber


IlseFerris:
Per SIR announcement dated [06/05/2023](https://www.dhssir.cty.dhs.state.mn.us/MAXIS/Lists/MAXIS%20Announcements/DispForm.aspx?ID=8021&RootFolder=%2FMAXIS%2FLists%2FMAXIS%20Announcements&Source=https%3A%2F%2Fwww%2Edhssir%2Ecty%2Edhs%2Estate%2Emn%2Eus%2FPages%2FDefault%2Easpx) there's a new INFO-based message that is available if any member on a case is coded as with the FSET code of 30 - Mandatory FSET Participant. Message is the following: POTENTIAL ABAWD MEMBER, REVIEW WREG PANEL, UPDATE AND APPROV&. 

DAIL scrubber already has several messages that support running the ABAWD FSET Exemption finder functionality. This would be another.

PR #1345, opened 2023-06-06
#1343 DAIL - DAIL SCRUBBER: Added new ABAWD message

IlseFerris:
Tested on a case in inquiry. Worked as expected.


C-Love (approved):
Looks good

PR #1357, opened 2023-06-16
850 abawd don't call it a come back

IlseFerris:
ABAWD scripts are ready to go. If this can get pulled in EARLY on Tuesday that would be ideal. There will be a Hot Topic coming out as well. All instructions updated on SPO. Tested scripts, but did not add closing issue documentation.


MDR22 (approved):
The changes look good. While testing, I noticed a minor typo in the notes/abawd-tracking-record script. Line 81 is error handling for the member number but it displays an error message of "Enter a valid footer year." Should be "Enter a valid member number."

PR #1368, opened 2023-06-26
1349 dail pick bug

IlseFerris:
I tested both ADMIN scripts, the TS doesn't need testing.


admin/dail-12-month-contact.vbs
MDR22, line 1: Only thought, should there be a changelog update for this script?

MDR22 (approved):
Looks good. Only comment related to changelog update for 12-month-contact script.

admin/dail-12-month-contact.vbs
IlseFerris: Whoops, forgot to add this one :) 
Good catch!

IlseFerris: @MDR22 fun fact, I did add the changelog, but I forgot to save it. Sigh...

Issue #1373, opened 2023-06-30
Training Case Creator HC REVW panel was updated


C-Love:
HC REVW panel needs the coordinates changed

IlseFerris:
Closing issue. Single user only actively using this script.

PR #1375, opened 2023-06-30
#1373 ADMIN - Training Case Creator - move HC REVW coordinates

C-Love:


IlseFerris (approved):
Looks good, and based on my experience with the REVIEW REPORT these are correct. Didn't test, but I approve thee!

PR #1386, opened 2023-07-07
1369 ex parte tikl support 

IlseFerris:
Partial issue pull request. @MDR22 take a look at the function non_actionable_dails to see which DAIL's are also excluded from assignment.

### DIFF
--- a/admin/task-based-dail-capture.vbs
+++ b/admin/task-based-dail-capture.vbs
@@ -247,6 +247,7 @@ For each worker in worker_array
 
             stats_counter = stats_counter + 1   'I increment thee
             Call non_actionable_dails(actionable_dail)   'Function to evaluate the DAIL messages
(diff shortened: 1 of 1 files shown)


MDR22:
> Partial issue pull request. @MDR22 take a look at the function non_actionable_dails to see which DAIL's are also excluded from assignment.

Helpful to see all of the excluded DAILs! Does this DAIL message need to get added to the non_actionable_dails function? Or is that unnecessary since it is noted as not actionable within this script?

IlseFerris:
That's correct. It wasn't added to the function since at this point we don't want it to be deleted, and that function is used in our scripts that delete as well.

IlseFerris:
Is this pull request a go @MDR22?

MDR22:
Yes, sorry! I commented but didn't approve. Looks good.

Issue #702, opened 2022-02-01
MEMB 'Access Denied' issues in scripts and functions


IlseFerris:
QI member AS, reported issue using the CAF script on a case where there is a member that is restricted with ACCESS DENIED coding. She also experienced inhibiting issues when using the CSR script. 

The CAF script issue appeared to be an issue where the MAXIS system wasn't reacting fast enough after the PF10 function was used, and the proceeding transmit didn't take. There was ACCESS DENIED handling, but needed an EMWaitReady for the system to catch up. 

The CSR script appears to have missing functionality in the function HH_member_custom_dialog - panel_read_from = "MEMB". 

The following scripts/functions to be reviewed for updates:

- [x] ADMIN - TESTING ADDR SHEL AND HEST FUNCTIONS
- [x] ACTIONS - FULL INTERVIEW: any members after the ACCESS DENIED members won't be added to array
- [x] DEU - ATR RECEIVED 
- [x] NOTES - CAF 
- [x] NOTES - INTERVIEW
- [x] UTILITIES - COMPLETE PHONE CAF - this has since been retired. No action needed.
- [ ] TEST SCRIPTS- NOTES COUNTY PAID MEDI B
- [x] UTILITIES - VA VERIFICATION REQUEST
- [ ] Function autofill_editbox_from_MAXIS - panel_read_from = "MEMB"
- [x] HH_member_custom_dialog(HH_member_array)

MDR22:
In reviewing all of the scripts, the following scripts have the access_denied_check without the EMWaitReady, but aren't on the #702 issue list - should I update these as well? 

- C:\MAXIS-scripts\admin\resolve-hc-eomc-in-mmis.vbs
- C:\MAXIS-scripts\notes\health-care-evaluation.vbs
- C:\MAXIS-scripts\admin\track-autoclose-overpayments.vbs
- C:\MAXIS-scripts\notes\mfip-orientation.vbs
- C:\MAXIS-scripts\Test scripts\Ilse\Restart scripts\BULK - restart IMIG.vbs
- C:\MAXIS-scripts\utilities\uc-verification-request.vbs

PR #1391, opened 2023-07-07
#702 memb 'access denied' issues in scripts and functions addl scripts

MDR22:
I updated the following scripts that were not initially identified in the issue but also contain access_denied_check. 

- resolve-hc-eomc-in-mmis
- admin\track-autoclose-overpayments
- notes\mfip-orientation
- Restart scripts\BULK - restart IMIG - I noted in commit but wasn't sure if this needed to be updated as it is in test scripts folder
- utilities\uc-verification-request


IlseFerris (approved):
Looks great, I approve thee!

Issue #1392, opened 2023-07-10
MA FIATER for MSA/GRH - bug if HC Income Estimate is not completed


C-Love:
If the income panel does not have all needed information on the HC Income Estimate, the script will error because this needs to be a number. 

We are going to stop the script run if this is missing.

C-Love:
This fix is working.

PR #1393, opened 2023-07-10
#1392 ACTIONS - MA Fiat MSA-GRH - end script if pop up is blank

C-Love:
Prioritized this fix since we have a great group of testers in the Ex Parte workers and we want to get everything in place. 

The goal is only to end the script if the pay frequency in the HC Inc Estimate is blank on a UNEA panel. 
Added a script run lowdown piece so the error reporting has more detail


IlseFerris (approved):
Makes sense, I approve thee!

MDR22 (approved):
Looks good to me - no changes.

Issue #1372, opened 2023-06-30
Create logic in the redirect chain for desert island


C-Love:
The Func Lib needs a logic statement for the right CLOS to call and we need to make sure the CLOS has a T Drive copy.

Desert Island broke on us.

C-Love:
This is all done!

PR #1394, opened 2023-07-10
#1372 correct desert island paths in Func Lib and CLOS

C-Love:
When the desert island failed, the redirects didn't work quite right. 

Most of the changes were needed in the local redirects for the desert island, but the Func Lib and CLOS are updated monthly from GH, so we needed to update the GH files with desert island paths.


IlseFerris (approved):
I didn't test, but saw that you had testing in the team email. 
I approve thee!

Issue #1241, opened 2023-04-12
BUG on CAF when calculating income into Exp Det for BUSI


C-Love:
The BUSI income was coming is as the GROSS amount, and not the NET amount.

Need to update the functionality to display the NET income.

C-Love:
This is updated and working, was well tested in the testing branch.

Issue #1203, opened 2023-03-13
NOTES - CAF - Have script select the programs to review


C-Love:
Update CAF so that using the Case Number, the script will select the programs that are being reviewed with the form.

Also a small bug in reading the interview case note for the interview date.

C-Love:
This is now working and deployed. DONE

PR #1395, opened 2023-07-10
#1203 & #1241 NOTES - CAF - change program selection and BUG on Exp Det BUSI income autofill

C-Love:
Updates to the CAF script to require CAF and Form dates and Interview dates be updated for the script to work correctly. This also removes the initial program selection as all should be reviewed.


IlseFerris (approved):
I've tested this previously and it has worked as anticipated. I approve thee!

Issue #1401, opened 2023-07-11
MA FIATer for MSA/GRH bug when NO asset panels


C-Love:
The script tries to do some very specific array/class things with information about assets and if none exists, it breaks down. 

Need to add a conditional statement to avoid doing array things if the array is empty. 

This is in one place in the script and just needs to be extended.

C-Love:
This is resolved.

PR #1402, opened 2023-07-11
#1401 - ACTIONS - MA FIAT for MSA-GRH - don't do asset things if no assets

C-Love:


IlseFerris (approved):
Didn't test but looks good. I approve thee!

PR #1406, opened 2023-07-12
#1203 NOTES - CAF better handling for finding if GRH is at ER

C-Love:
There is logic issues in figuring out if we are at SR or ER for the programs with SRs. 

Currently GRH is showing an issue because it should allow an ER to process with no interview, but it didn't even think this was an ER BECAUSE the SR and ER are the same date.

More logic to ensure that we know what process GRH is at.


IlseFerris (commented):
Looks good, I approve thee!

Issue #1410, opened 2023-07-14
Case Number entered into SQL table in invalid format


C-Love:
There was an ES Workflow issue because a case was entered into a SQL table looking a mess ('0 XXXXXX')

Spaces are not valid and the case number needs to be 8 digits. 

Pretty sure this is a trim issue. 

Going to check the scripts that update this table.

C-Love:
The case that caused the error had Application Received run on it and the err handling in the dialog doesn't use the validate_MAXIS_Case_number so the case number was never trimmed. 

Pretty sure this was the cause of the issue, but for safety, we should just trim the variable right before we create the SQL case number with the leading 0s.

C-Love:
This is resolved and hopefully we never cause this error again.

PR #1411, opened 2023-07-14
#1410 Trim Case Number for SQL to be in error

C-Love:
I think the main issue was the MAXIS_Case_Number error handling in the case number dialog in Application Received, BUT might as well solidify everything that updates the cases pending table.


IlseFerris (approved):
I approve thee!

PR #1414, opened 2023-07-17
#1203 - NOTES - CAF - read PROG if REVW is T - make app if app date after ER

C-Love:


IlseFerris (approved):
This looks like it will work. I approve thee!

Issue #1422, opened 2023-07-20
DAIL - BNDX Scrubber: DAIL deletion functional not working


IlseFerris:
JS, BZS tester reported that the DAIL deletion function was not working and that the script keeps running without taking the applicable deletion action.

IlseFerris:
No negative reports provided by users. Closing issue.

PR #1423, opened 2023-07-20
#1422 DAIL - BNDX Scrubber: fixed bug in delete function

IlseFerris:
Added changelog
Added reading and trimming full DAIL message and measurement dail message. Added warning message handling to delete messages not in workers number.


C-Love (approved):
I see no issues here. I did not run it.

Issue #1425, opened 2023-07-21
Add email output to Resolve HC EOMC for QI 


C-Love:
QI needs to know when the EOMC list is done, this will email the QI team email when the script run is complete.

C-Love:
Once #1409 is completed we will need to realign the email functionality to use the actual function.

This also needs one more month of testing before we can close because the tables and autosave were not tested on a live run.

C-Love:
This is completed.

PR #1426, opened 2023-07-21
#1425 ADMIN - Resolve HC EOMC - add email output to QI

C-Love:
This will be the first part of this resolution until we can get it aligned with the FuncLib.


IlseFerris (approved):
Looks good. I approve thee!

Issue #1418, opened 2023-07-18
Script Usage Output - update the set of date and time


C-Love:
Currently the function to save the script run details sets the date and time AFTER the closing messagebox is cleared (user presses OK). This creates a delay between the actual script run date/time and what is recorded in the statistics. 

We need to set the date and time to a variable prior to the display of the messagebox. These variables will then be passed through to the database after the messagbox is clear. This will provide more accurate stats for script usage.

C-Love:
This is done! I didn't break it.

Issue #1228, opened 2023-04-06
Function determine_program_and_case_status_from_CASE_CURR: REIN status has dates


IlseFerris:
CAF testing/script low down shows GRH date is cut off. Review and update outputs.

C-Love:
This is actually an issue with ALL programs on REIN. The MM/YY is listed next to the REIN status. We need to reset that status to exclude the MM YY from the prog_status variable.

C-Love:
This appears to be functioning - AND we didn't break the function or the func lib.

PR #1437, opened 2023-07-27
#1418 #1228 FuncLib updates to statistics and determine_prog_and_case_status

C-Love:
Stats update:
We need to set the script date and time for when the script actually stops and not when the worker presses 'OK' on the end message box. This is facilitated by setting the date and time to variables that are defined before the display of the messagebox. Then, it doesn't matter WHEN the information is written to the data table, it is set from the end of the script run functionality. 

Update to determine_program_and_case_status_from_case_curr
There was an issue when the program status was read from CASE/CURR - there was extra information in the variable. I could only find this on REIN status, this is because the MM YY are listed next to the REIN status on CASE/CURR
image
The fix just resets the status variable in the function when 'left, 4' is REIN. This way only 'REIN' will pass back out of the function.


IlseFerris (approved):
Looks good. I approve thee!

Issue #1378, opened 2023-07-01
NOTES - CAF: Add option to add DHS-2128 form for Housing Support renewals


IlseFerris:
[RECERTIFICATION PROCESS (state.mn.us)](https://www.dhs.state.mn.us/main/idcplg?IdcService=GET_DYNAMIC_CONVERSION&RevisionSelectionMethod=LatestReleased&dDocName=cm_000906) - Policy for using DHS-2128 LTC renewal for Housing Support program renewals.

C-Love:
This is complicated because this form cannot be used for all programs. We need to create some clear logic and understanding of how this form operates and what we can do to support it.

In discuss at next meeting to try to make a plan for answering the questions:

- What happens when the form is received and there is an unsupported program?
- Is it only a problem if the program is pending or at renewal?
- What is the scope of this issue? How many of these forms are we seeing?

C-Love:
Plan 1 - to get us started.
1. Add the form
2. IF any other cash program (other than GRH) or SNAP is active and at renewal for this form selected the script will stop
3. Message will alert the worker that policy on using this form for other recert programs is a problem and PQ is needed. 

This will be a simple update but not supportive of confusing situations.

C-Love:
This is in place, no errors reported. Done.

Issue #1352, opened 2023-06-08
Redirect from CAF to HC Eval if case is ONLY HC


C-Love:
CAF is program specific and there is currently no HC support. Need to redirect to the right script.

PR #1438, opened 2023-07-28
#1378 & #1352 - NOTES - CAF - adding DHS-2128 & Redirect to HC Eval

C-Love:
There are 2 changes here to support handling in the CAF.

Added Renewal for MA-LTC people (2128) form option - but the script will stop if programs other than GRH have a recert.
If the CAF cannot continue, and the case has HC and a form that can be used for HC is selected, the option to run HC Eval is offered.


IlseFerris (approved):
Looks good. I approve thee!

MDR22 (approved):
No edits - looks good to me!

Issue #1447, opened 2023-08-17
Digital Drop Box is Sundowning


IlseFerris:
I just met with SF, our IT business partner, and the digital drop box is sundowning at the end of the year. The only active script that has this reference is ADMIN - INDIVIDUAL RECERTIFICATION NOTICES. 

- [x] Remove verbiage that points to digital dropbox (hhsews@hennepin.us) and replace with verbiage in screenshot of Appointment Notices in Review Report/Application Received.
- [x] Verify it is using English and not another language. Verified that 'English' is hard coded into the script.

C-Love:
This is completed.

PR #1456, opened 2023-08-28
#1447 removed dropbox ref

MeganGeissler:
Removed digital dropbox (hhsews@hennepin.us) and replaced with updated verbiage. Verified English is hard coded into script. Did not add to change log since it was such a minor change. Let me know if you feel differently.


IlseFerris (approved):
Looks good. I approve thee!

PR #1459, opened 2023-08-28
1036 create make appointment &nomi functions

MeganGeissler:
2 separate commits: 
- Appointment notice
- NOMI

Let me know what feedback you have if any!

### DIFF
--- a/MASTER FUNCTIONS LIBRARY.vbs
+++ b/MASTER FUNCTIONS LIBRARY.vbs
@@ -5364,7 +5364,7 @@ function create_appointment_letter_notice(application_date, interview_date, last
 	CALL write_variable_in_SPEC_MEMO(" - 1001 Plymouth Ave N Minneapolis 55411")
 	CALL write_variable_in_SPEC_MEMO(" - 2215 East Lake Street Minneapolis 55407")
 	CALL write_variable_in_SPEC_MEMO(" (Hours are 8 - 4:30 Monday - Friday)")
-	CALL function digital_experience()
+	CALL digital_experience
 	CALL write_variable_in_SPEC_MEMO("Domestic violence brochures are available at this website: https://edocs.dhs.state.mn.us/lfserver/Public/DHS-3477-ENG. You can always request a paper copy via phone.")
 	PF4
 end function
@@ -5472,7 +5472,7 @@ function create_NOMI (application_date, appt_date, last_contact_day)
 	CALL write_variable_in_SPEC_MEMO(" (Hours are 8 - 4:30 Monday - Friday)")
 	CALL write_variable_in_SPEC_MEMO(" More detail can be found at hennepin.us/economic-supports")
 	CALL write_variable_in_SPEC_MEMO("")
-	CALL function digital_experience ()
+	CALL digital_experience
 	PF4
 end function
 
--- a/MASTER FUNCTIONS LIBRARY.vbs
+++ b/MASTER FUNCTIONS LIBRARY.vbs
@@ -6410,7 +6410,6 @@ function digital_experience()
 	CALL write_variable_in_SPEC_MEMO("- Use InfoKeep online at infokeep.hennepin.us to upload directly to your case.")
 	CALL write_variable_in_SPEC_MEMO("- You can submit documents online at MNBenefits.mn.gov.")
 	CALL write_variable_in_SPEC_MEMO("- Mail, Fax, or Drop Boxes at service centers.")
-	Call write_variable_in_SPEC_MEMO(" ")
 end function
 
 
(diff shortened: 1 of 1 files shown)


IlseFerris (approved):
Looks good to me. @C-Love I'm going to add you to the review since you use these the most in your code.

MeganGeissler:
@IlseFerris  & @C-Love - Revised code to eliminate error. Ready for review!

IlseFerris (approved):
I approve thee, again!

C-Love (changes requested):
It looks like these are both for applications only. We have separate phrasing for recertification for both the appointment notice and the NOMI. You can find these in ADMIN - Review Report line 2521 and line 3348. 

These have different parameter needs, so we might want different functions but perhaps it makes more sense to have a parameter make the distinction. This will potentially create 4 functions, which is fine but something to keep an eye on. 

Additionally, we might want to consider future planning for other languages. Adding a language parameter (even if it isn't functional) allows us to add that without having to update each function call. This is certainly optional.

MeganGeissler:
@C-Love - I've created 4 functions total to account for recertification for appointment and NOMI. Please review to see if I have the correct coding for all 4 functions! :) Thanks for your review.

MASTER FUNCTIONS LIBRARY.vbs
```diff
@@ -5340,6 +5340,73 @@ function convert_digit_to_excel_column(col_in_excel)
...
+	CALL write_variable_in_SPEC_MEMO(" - 1001 Plymouth Ave N Minneapolis 55411")
+	CALL write_variable_in_SPEC_MEMO(" - 2215 East Lake Street Minneapolis 55407")
+	CALL write_variable_in_SPEC_MEMO(" (Hours are 8 - 4:30 Monday - Friday)")
+	CALL digital_experience
+	CALL write_variable_in_SPEC_MEMO("Domestic violence brochures are available at this website: https://edocs.dhs.state.mn.us/lfserver/Public/DHS-3477-ENG. You can always request a paper copy via phone.")
+	PF4
+end function
+
+
+function create_appointment_letter_notice_recertification(programs,CM_plus_1_mo,CM_plus_1_yr, intvw_programs, interview_end_date, last_day_of_recert)
```
C-Love, line 5373: CM_plus_1_mo,CM_plus_1_yr - these are globally defined, and you do not need to pass them through as arguments. You can remove them from the function parameters. Leave them in place within the function and they will work just fine.

MASTER FUNCTIONS LIBRARY.vbs
```diff
@@ -5340,6 +5340,73 @@ function convert_digit_to_excel_column(col_in_excel)
...
+	CALL write_variable_in_SPEC_MEMO("** we receive your paperwork and complete the interview. **")
+	CALL write_variable_in_SPEC_MEMO("")
+	CALL write_variable_in_SPEC_MEMO("All interviews are completed via phone. If you do not have a phone, go to one of our Digital Access Spaces at any Hennepin County Library or Service Center. No processing, no interviews are completed at these sites. Some Options:")
+	CALL write_variable_in_SPEC_MEMO(" - 7051 Brooklyn Blvd Brooklyn Center 55429")
+	CALL write_variable_in_SPEC_MEMO(" - 1011 1st St S Hopkins 55343")
+	CALL write_variable_in_SPEC_MEMO(" - 1001 Plymouth Ave N Minneapolis 55411")
+	CALL write_variable_in_SPEC_MEMO(" - 2215 East Lake Street Minneapolis 55407")
+	CALL write_variable_in_SPEC_MEMO(" (Hours are 8 - 4:30 Monday - Friday)")
+	Call digital_experience
+	CALL write_variable_in_SPEC_MEMO("Domestic violence brochures are available at this website: https://edocs.dhs.state.mn.us/lfserver/Public/DHS-3477-ENG. You can always request a paper copy via phone.")
```
C-Love, line 5406: I am a little concerned here because the digital experience function includes a space after the information and this notice is very full and cannot fit if there is a space between the digital experience content and the DV brochure detail.

C-Love (changes requested):
I put the comments in line. There are some globally defined variables that you do not need to add as parameters and the MEMOs have some really strict space handling needs that I think will be an issue.

MASTER FUNCTIONS LIBRARY.vbs
MeganGeissler: Makes sense, thank you! Clean is key.

MASTER FUNCTIONS LIBRARY.vbs
MeganGeissler: Committed separately and tagged digital experience issue to indicate the space has been removed to accommodate its use with other sections of memo.

MeganGeissler (commented):
Revised code to include spaces where available.

PR #1480, opened 2023-09-11
1458 replace ind digital experience with function

MeganGeissler:
Updated individual coding with function for digital experience. 
NOTICES-ADD WCOM leveraged add_words_to_message so I simply updated the words to match the function built.  Let me know if you want something different here.


IlseFerris (approved):
I approve thee!

PR #1481, opened 2023-09-11
1036 apply function appointment notice

MeganGeissler:


MeganGeissler:
Applied function to 3 scripts.

IlseFerris (changes requested):
@MeganGeissler will you add changelog's to these script files? Once that's done, go ahead and pull these changes in. After digital experience pull request though :)

Issue #1452, opened 2023-08-24
BUG - Spelling Error in HC Eval for MA-BC Presumptive Eligibility


C-Love:
Review the CASE/NOTE output for the MA BC Presumptive Eligibility, there are spelling errors.

8/1/2023 SAGE Enrollment Form : Complete- Ineligible ( Could this aspect be added when denied)  
 * Form Recvd: SAGE Enrollment Form (MA/BC PE Only)                           
 * Date Recvd: 8/1/2023                                                       
 * Temporary Medical Assistance Authorization (DHS-3525B) recvd on: 8/1/2023  
 ========================== PERSON DETAILS ==========================         
 MEMB 01 - MARIA DIAZ approved for MA-BC Presumptive Care.                    
   Presumptive Care to be approved for 08/23 and 09/23.                       
 * Method X Budget - no Income or Assets needed for Presumptive **_Eligbility_**.   
* Identity verified using medical document - SAGE Enrollment Form .          
 * Citizenship and Immigration information are not requested or required.     
 ============================== NOTES ===============================         
 * Notes: Client is not eligible for Breast and Cervical Cancer MA basis of   
          presumptive care due to being undocumented and active on EMA       
          insurance.   
  TIKL to close MA-BC **_Preseumtive_** Eligibility.   
 ---                                             
 M. Socha SAGE/LTC+

Issue #1359, opened 2023-06-22
NOTES - HC Eval - update to try and identify if case is recert processing to autofil the status


C-Love:
To support accurate selection of HC process, we should read STAT/REVW to identify if a case is at REVW for the run of HC Eval. 

We need to read STAT/REVW for each footer month up to CM-4 (though only back to 07/23) and read the case based code for HC REVW. If a code exists, we will default all the HC members to recert in the status that is processing.

This is an aim to prevent the wrong process selection to trigger the 'return to standard policy' CASE NOTE.

PR #1483, opened 2023-09-13
#1452 & #1359 HC EVAL - Fix typos for presumptive eligibility and add functionality to determine if case is a recertification

MDR22:
#1452 - Fixed a few typos related to presumptive eligibility

#1359 - Added functionality to navigate to STAT/REVW for CM + 1 through, and including, CM - 4 (but not past 07/23) to determine if a HC status code is present. For instances where the HC status code is present, script will default HC eval process to "Recertification".

### DIFF
--- a/notes/health-care-evaluation.vbs
+++ b/notes/health-care-evaluation.vbs
@@ -4772,12 +4772,12 @@ If HC_form_name = "SAGE Enrollment Form (MA/BC PE Only)" or HC_form_name = "Scre
 				EditBox 10, 190, 365, 15, ma_bc_tikls
 				OkButton 270, 215, 50, 15
 				CancelButton 325, 215, 50, 15
-				Text 10, 15, 185, 10, "Select the Person with MA/BC Presumptive Eligiubility:"
+				Text 10, 15, 185, 10, "Select the Person with MA/BC Presumptive Eligibility:"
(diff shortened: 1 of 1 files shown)


notes/health-care-evaluation.vbs
```diff
@@ -4925,6 +4985,7 @@ FOR i = 0 to total_clients
 				If HEALTH_CARE_MEMBERS(ref_numb_const, hc_memb) = HH_memb Then
 					HEALTH_CARE_MEMBERS(show_hc_detail_const, hc_memb) = True
 					HEALTH_CARE_MEMBERS(HC_major_prog_const, hc_memb) = "MA"
+					If hc_status_code_past_4_months = False Then HEALTH_CARE_MEMBERS(HC_eval_process_const, hc_memb) = "Recertification"
```
C-Love, line 4988: I think this is supposed to be "If hc_status_code_past_4_months = True Then ". The logic to get the boolean set correctly, but the 'Recertification' does not come up.

C-Love (changes requested):
The logic is solid and easy to follow. I think you just have the wrong Boolean value set in setting the array parameter for the person.

notes/health-care-evaluation.vbs
MDR22: Yes, you are 100% correct. Fixed that.

C-Love (approved):
Looks great!

Issue #404, opened 2021-03-19
Earned Income Budgeting - Add Missing Check YTD Calculator


C-Love:
Add a button/functionality to calculate missing checks in a series using the TYD information.

C-Love:
This seems to be resolved and is working.

PR #1492, opened 2023-09-18
#404 ACTIONS - EIB - YTD Calculator - bug with blank check

C-Love:
Response to AGK email in team email. Bug when there is a blank check entered.


IlseFerris (approved):
Makes sense to me, I approve thee!

Issue #1034, opened 2022-11-02
New Script to update the HSR list for ADMIN - Expedited Determination Report


C-Love:
The script ADMIN - Expedited Determination Report accesses a file that has a list of HSRs and HSSs maintained within it to identify workers.

This list needs to be kept up to date. 

We are working on access to 'InfoStore' that will allow us to use a script to pull this updated date from the table and match the Excel that ADMIN - Exp Det Report uses. 

This will be run regularly as a part of our report maintenance.

C-Love:
Currently waiting on access.

C-Love:
Functionality is in place for this to operate but we need to get access to the data table for the OS person who runs this script.

C-Love:
To discuss:
There is a branch for these updates with 2 primary functionality parts:
1. Remove the HSS list sheet from the way the script finds user information and change the functionality to find user information directly from the SQL data table. This functionality works but our primary user of this script (KR) does not have access to this data table. 
2. Create a reset functionality to update the HSS list using the SQL data table. This functionality can be used by the script team to update the HSS list for the current script functionality. This will improve the way we update the Excel sheet that supports Expedited Determination Spot Checks processing. Anytime we get a missing HSR report, this can be used. 

The second functionality should not be needed once KR gets access to the data table, since the script will access the information directly. 

Questions to decide what comes next:
1. When do we expect KR to have access?
2. Should we 'disable' the first functionality and merge the branch? This will make putting the update in place very quick and will keep us from having to maintain a branch.

C-Love:
Answers - no idea about access for KR. 

So yes, I will 'disable' the new functionality in the standard run and put the old functionality back in place and merge this branch once the functionality is able to be run with current access.

C-Love:
These updates seem to be working - no errors reported.

PR #1496, opened 2023-09-22
#1034 ADMIN - Expedited Determination Report - using SQL table

C-Love:
Updates to ADMIN - Expedited Determination Report:
1. Change the reading of user information from Excel to SQL - BUT put it on HOLD while we work on access to the data table
2. Adding the ability to completely reset the HSS list in Excel since this will need to be updated until we can directly access the data table. 

This script is used by KR almost exclusively.


IlseFerris (approved):
Looking good. I approve thee!

Issue #1501, opened 2023-09-29
ADDR Panel - the phone numbers moved 


C-Love:
There was an update overnight in which the phone numbers moved on the ADDR panel to make room for a new system entered field about the source of the text authorization.

Currently it appears this fix should be able to be managed in the access_ADDR function in the functions library. Will continue to look for other places phone numbers are pulled without using the function.

C-Love:
Look we fixed it.

PR #1503, opened 2023-09-29
#1501 - ADMIN - Training Case Creator - ADDR phone number moved

C-Love:
Updates to Training Case Creator with the new phone number positioning on ADDR. 

Tested on a scenario and it worked well. (TRN - 329796)


IlseFerris (approved):
I approve thee!

Issue #1465, opened 2023-08-29
UTILITIES - All Scripts - add scriptwriter support to evaluate policy_references 


C-Love:
It would be helpful to create an error in All Scripts if any item listed in policy_references has the wrong number of items in the array for each policy. Done with enough information, this would display which script and which policy had the wrong number of array items, there should always only be 0, 1, 2 for array options, a UBound of 2. This prevents us from breaking our eyes trying to make sure spaces and underscores are in the right spot.

C-Love:
This is mostly used internally and changes appear to be working.

PR #1504, opened 2023-09-29
#1465 UTILITIES - All Scripts - add testing block of policy references

C-Love:
If we update CLOS policy.reference item with a new policy, there are specific requirements that are fussy. This adds a check and message box to All Scripts to help support testing of these updates.


IlseFerris (approved):
Lookin' good. I approve thee!

PR #1506, opened 2023-09-29
1413 1281 bulk poli temp

IlseFerris:
Complete list of scripts will be pulled in separately. Instructions will be held locally, created this week by Jake and I.


MDR22 (commented):
I read through the code and it looks good. For the initial dialogs, I wasn't sure about the situations where the dialogs would come up due to the worker code _not_ being X127. I was still getting the error code for line 14 so wasn't able to fully run the script. I will try again later today.

IlseFerris:
@MDR22 I updated the CLoS, and I created a new poli temp file. Do you want to test it out again? When the script asks you the county code, instead of entering 27, you can enter 62 or some other county.

MDR22 (commented):
I made a comment in the functions library code block on line 12. I am still getting the incorrect parameter error on line 14. I think it has to do with the if statement on line 12 related to use_master_branch.

MDR22 (approved):
That change fixed the error. I tested it and it worked great for me!

Issue #1508, opened 2023-10-03
Add new scriptwriter to redirect files to support local global variables


courtrightd:
It's me! I'm the new scriptwriter!!!

PR #1509, opened 2023-10-03
#1508 add new scriptwriter to redirect files

courtrightd:
Simply adding my username into all the menu redirects for local testing purposes.  
All buttons tested and working on my end.


IlseFerris (changes requested):
redirect-mc-main-menu is missing. Throw that in, and then go ahead and pull it in. I approve thee!

Issue #1514, opened 2023-10-06
Test MNIT Script for initial PERS Search functionality


C-Love:
MNIT has provided a script to search by specifc person search options that need testing. 

We will be implementing this script for a testing period of two weeks. Additionally this script has initial functionality to parse the xml file. 

Issue to create testing file for CA as well as saving the file in misc for future code hunting.

C-Love:
Testing of this is completed. The script is in place.

PR #1515, opened 2023-10-06
#1514 MNIT Person Search Script add for testing

C-Love:
This script was developed by MNIT and does not necessarily follow all of our standards/conventions. So there will be things that do not look quite right, and that's fine for this. 

I have changed only the parts necessary to make the script run in our infrastructure and then some supportive measures for our testers to reduce the burden of testing.

### DIFF
--- a/case-assignment/person-search.vbs
+++ b/case-assignment/person-search.vbs
@@ -1,0 +1,247 @@
+'Required for statistical purposes==========================================================================================
+name_of_script = "CA - PERSON SEARCH.vbs"
+start_time = timer
+STATS_counter = 1                          'sets the stats counter at one
+STATS_manualtime = 10                      'manual run time in seconds
+STATS_denomination = "I"                   'I is for Item - based on search criteria
+'END OF stats block=========================================================================================================
+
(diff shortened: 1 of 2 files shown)


case-assignment/person-search.vbs
```diff
@@ -0,0 +1,247 @@
...
+  ButtonGroup ButtonPressed
+    CancelButton 60, 50, 45, 20
+EndDialog
+
+
+DO
+			err_msg = ""
+			Dialog Dialog1
+			If ButtonPressed = 0 then Stopscript
+			If Search_type = "Select" THEN err_msg = err_msg & vbCr & "Please select SSN or Name and DOB."
```
MeganGeissler, line 58: Should this error handling also include verbiage for Last Name and First Name since this is a valid option from your dropdown?

C-Love: This is part of the script from MNIT that I am letting run as is. I would make a lot of changes here if it were up to me.

MeganGeissler: Touche

MeganGeissler (approved):
Looks good to me as far as I can tell!

C-Love:
Great! Thanks @MeganGeissler

Issue #1512, opened 2023-10-05
NOTES - SPONSOR INCOME: Update to make dynamic with yearly updates


IlseFerris:
This script needs yearly updates in October due to using 130% for calculations. We have since added the function determine_130%_of_FPG to the FuncLib, but this script does not use this yet. It should since this will make it more dynamic, and make it so we don't need to actually update it yearly.

PR #1516, opened 2023-10-09
#1512 sponsor income uses function to determine 130% FPG

courtrightd:
This required adding footer month to dialog to allow it to calculate the 130% based on footer month.


IlseFerris (approved):
Yay, it works! I approve thee!

PR #1519, opened 2023-10-09
1449 notes hc transition person record number

MeganGeissler:
Updated NOTES-HC transition to include person record number if METS case number is not available. 
Updated CLOS to include documentation.


IlseFerris (approved):
Looking good, I approve thee!

Issue #1491, opened 2023-09-18
Update to NOTES - Appeals - Docket Number is global


C-Love:
Request for R H-S around the appeals script. The option for 'Resolution' does not have a field for a docket number. Docket number should be required for all options in appeals. 

The docket number should be entered in the initial dialog and mandatory (it is alright to allow for 'N/A' as the entry for docket number). This way the docket number is available for all options in the appeals run.

PR #1523, opened 2023-10-09
#1491 NOTES - Appeals docket number is global

courtrightd:
Added docket number to the initial dialog, and error handling to require the user enter the number for most scenarios on the script. All scenarios now case note the docket number.

### DIFF
--- a/notes/appeals.vbs
+++ b/notes/appeals.vbs
@@ -71,16 +71,18 @@ CALL MAXIS_case_number_finder(MAXIS_case_number)
 
 '-------------------------------------------------------------------------------------------------DIALOG
 Dialog1 = "" 'Blanking out previous dialog detail
-BeginDialog Dialog1, 0, 0, 176, 85, "Appeals"
+BeginDialog Dialog1, 0, 0, 176, 105, "Appeals"
   EditBox 75, 5, 45, 15, MAXIS_case_number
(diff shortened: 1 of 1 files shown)


notes/appeals.vbs
```diff
@@ -90,6 +92,8 @@ Do
 		Dialog Dialog1
 		cancel_without_confirmation
       	Call validate_MAXIS_case_number(err_msg, "*")
+        IF docket_number = "" then err_msg = err_msg & vbNewLine & "* Please enter a docket number or enter N/A if unknown."
+        IF appeal_actions <> "Received" AND appeal_actions <> "Summary Completed" AND isnumeric(docket_number) = False then err_msg = err_msg & vbNewLine & "* You have selected an action that should correspond to an existing docket number. Please enter a numeric docket number."
```
IlseFerris, line 96: Line 96 - I think this can be removed now since docket number is now universal.

IlseFerris (changes requested):
Ok just the one change, and can you add a changelog? Then I approve thee!

courtrightd:
Added changelog.

notes/appeals.vbs
courtrightd: This is needed to allow for script actions that do not require docket number.

Issue #675, opened 2021-12-15
UTILITIES Menu is too large and is causing a Program Exception Error


C-Love:
The symptom is showing up in UC Verification Request.

Too many buttons, adding a new dialog with a different variable designation is causing the overload. 

We need to update the menu to NOT have so many items.

IlseFerris:
Review the category buttons (TOOL, POLICY, MAXIS). Have the top tools that are not on the power pad directly on the 1st/main page which include the following scripts: 
- Calculate rate 2 units
- Display benefits
- Insert MBI from MMIS
- POLI TEMP to Word
- Search Case Notes
- UC Verification Request
- Update Check dates 
- VA Verification Request

C-Love:
Additional notes in LAN in the Project folder.

C-Love:
This is working with new layout. No errors reported.

PR #1524, opened 2023-10-09
#675 UTILITIES - new subcategories  and it's not longer too big

C-Love:
Removed 'ALL' option added 'TOP' and 'REQUESTS'


IlseFerris (approved):
Ohhh I'm so excited about this one. I approve thee!

Issue #1495, opened 2023-09-21
NOTES-CAF In ABPS panel, create handling for names that have more than 1 space between names


MeganGeissler:
If there are two spaces between a name, the user will receive a "Script Error at Line 1, Column 1: Invalid procedure call or argument: 'right' error" in the ABPS panel.  Example: Charles_ _Jr

Current workaround: Manually delete extra spaces between the names to only have one space then script will run successfully. 

Need handling to look for and remove extra spaces to eliminate this error for future users. Desired format:  Charles_Jr

PR #1526, opened 2023-10-09
#1495 Add handling for names with extra spaces

MDR22:
Add handling for instances on ABPS panel where there are extra spaces between multiple last names


C-Love:
Hey @MDR22 - the function has been added to the CAF script here - is there a reason it isn't just updated in the functions library.

MDR22:
> Hey @MDR22 - the function has been added to the CAF script here - is there a reason it isn't just updated in the functions library.

That was a mistake. I only intended to make the update in the functions library. The edits to the CAF script were part of testing. I will fix that!

MDR22:
@C-Love I updated the pull request to only include update to function in functions library.

C-Love:
Awesome! This is also great because I was afraid we were going to have a merge conflict.

C-Love (approved):
Looks good!

Issue #1443, opened 2023-08-03
NOTES - CAF - Remove second form option if a second process is untracked


C-Love:
There was an issue reported where a case had a GRH ER for 09/23 but also had SNAP pended from 6/8/23. This was an MFIP to SNAP transition process and no form was associated. Both programs came up, but SNAP was allowed to be 'Untracked', when SNAP was untracked, the dialog still had a second form option and would not let it be blank. We need to remove that second form option if programs are untracked and there is only one form date once the untrack is selected. 

We also need to realign the dates for the form and interview to the selected options if it changes
![image](https://github.com/Hennepin-County/MAXIS-scripts/assets/13418322/d1d099f9-db14-4df7-a8b4-3d24f72959ea)

![image](https://github.com/Hennepin-County/MAXIS-scripts/assets/13418322/8eeb1150-4788-46c7-a097-1505fb89d717)

C-Love:
No errors reported. This didn't appear to break the CAF.

PR #1528, opened 2023-10-10
#1443 NOTES - CAF - removing multiple dates if a program is 'unchecked' that caused the multiple dates

C-Love:
There is some testing info on line 6184 that I will remove before pulling in.

This is a unique situation and we need to 'force' it to test. A case number is also listed there that will work to test. You should be able to 'untrack the EMER' and if that is done, the APPL date will not be considered and the second form will go away.


C-Love:
Good call @courtrightd  - I will add one as I pull the testing line out.

Issue #1529, opened 2023-10-10
MHC Scripts - update for annual changes


C-Love:
The MHC Enrollment script updates MMIS for enrollment actions. Every year we need to make changes to support processing of these cases. 

- [x] Update Open Enrollment to be for 2024
- [x] Update the capitation dates for 2024 when published

Sometimes there are plan changes, but we have not heard of any plan changes at this time. 

Capitation dates are monitored here: https://mn.gov/dhs/partners-and-providers/policies-procedures/minnesota-health-care-programs/provider/mcos/contract-information-forms-and-resources/health-plan-systems-workgroup/ 

This will be completed in Q4 as the information is available.

C-Love:
All updates completed.

PR #1530, opened 2023-10-10
#1529 MHC - Enrollment - move to Open Enrollment for 2024

C-Love:
Just a quick change to 2024 open enrollment


MDR22 (approved):
I saw that in past years you updated the changelog, that would be my only suggestion. Otherwise, looks good to me.

Issue #1531, opened 2023-10-12
DAIL Scrubber bug - cannot run on a DAIL in certain multi-member circumstances


courtrightd:
Scrubber is missing reading the top line when a case has more than one page of DAILs but includes a DAIL for 2nd member that ends up on top line. 
Also triggers a type_mismatch when running to the end of the DAIL list searching for the DAIL.

PR #1532, opened 2023-10-12
#1531 dail scrubber multi member bug

courtrightd:
Corrects issues so that scrubber can find the correct DAIL without getting into an endless loop or triggering a mismatch.
To test, just make sure it opens a script on the DAIL.
Case number that was problem is in our inbox.


C-Love:
@courtrightd  - is this to resolve the bug reported yesterday?

courtrightd:
> @courtrightd - is this to resolve the bug reported yesterday?
That is correct

C-Love (approved):
Looks good. 

Also, yikes, this was a mess.

Issue #1233, opened 2023-04-10
LTC Approval-Based Scripts to Retire/Update w/ Eligibilty Summary Core Functionality Release


IlseFerris:
With core functionality for health care programs completed and tested, let's look at retiring/updating elig based scripts for Health Care. These will include:
- [x] DAIL - COLA Review and Approve
- [x] NOTES - LTC - COLA Summary
- [x] NOTES - LTC - Intake Approval
- [x] NOTES - LTC - MA Approval

C-Love:
Evaluation of these scripts indicates we should do the following:

1. Update DAIL - COLA Review and Approve to be more of a COLA income summary to document the income information. #1518 
2. Remove all 3 LTC scripts
3. LTC Intake Approval has some functionality that should be a part of HC Evaluation and will need to be supported before the script can be removed #1517 

None of these scripts will have support to document approvals directly any longer. This also supports the work in #1166 as these scripts have some weird check_for_MAXIS calls - but they won't be a problem once the scripts are gone.

C-Love:
This is done, these have been retired.

PR #1537, opened 2023-10-16
#1233 Retire LTC COLA Summary and LTC MA Appproval and DAIL COLA Review and Approve

C-Love:
These approval notes are now supported in Eligibility Summary and should not be processed with a different script.


IlseFerris (approved):
I approve thee!

PR #1544, opened 2023-10-20
1543 Banked Months Updater bug fix for ineligible members

courtrightd:
This fix corrects a reported issue where members without a WREG panel were getting variables carried over from previous member. Moved the variable resets outside of if statement to resolve. 
Added case note option to note no WREG panel when listing all members' WREG status in case note.


IlseFerris (approved):
Looks good. I approve thee!

Issue #1517, opened 2023-10-09
Add LTC/Waiver Initial assessment dialog to HC Evaluation


C-Love:
The processing for LTC and Waiver cases upon their initial application for LTC/Waiver has some additional requirements and processing steps/details that need to be documented. 

This includes:

- Lookback period
- Baseline date
- Start of MA
- Start of LTC
- LTCC indicator
- Indicate if client is in the community or not
- CFR (this might already exist)
- Is there LTC partnership
- Space for Annuity LTC PRB
- Home equity limit
- Transfer information
- Forms - 5181, 1503, 3050, 3203 (lien), asset transfer letter, managed care letter

This information will be available in HC Eval when the worker indicates that this is a new request for LTC/Waiver. This functionality is currently supported in NOTES - LTC Intake Approval. We will be retiring that script once the supports are in HC Eval. 

This will create a stand-along CASE/NOTE regarding the initial LTC/Waiver Evaluation and Assessment. The Approval of any MA will be supported by Eligibility Summary. 

Connected to work on #1233

C-Love:
This appears to be working, no error reports have come in.

PR #1550, opened 2023-10-27
#1517 & #1233 Add LTC Intake functionality to HC Eval

C-Love:
In retiring the LTC approval scripts, we found functionality that is beneficial in creating a NOTE about LTC Intake with specific fields. 

These are being added to HC Eval.


C-Love:
I know that there hasn't been reviews of this pull yet. I am going to pull it in now so if there are any obvious issues, I can catch them while I am still here.

Issue #1549, opened 2023-10-27
NOTES - Vendor does not loop through search dialog correctly.


courtrightd:
When pressing the vendor search button, the script jumps to the 2nd dialog after searching, regardless of what the user does in the search dialog. It should loop back to the initial dialog, but err_msg value is reset.
This should also fix this script for #1166

Issue #1166, opened 2023-02-16
Audit check_for_MAXIS in the project and align with Dialog1 conventions 


C-Love:
The check_for_MAXIS function has a different dialog name than the standard Dialog1. We would like to get this function aligned. First we need to audit the whole project to ensure that check_for_MAXIS is not inside any dialog loops (because this would cause the deletion of dialog definition for that loop). 

Make sure that all the calls for this function are outside of dialog loops.
Once these are done, update the function to use Dialog1 as the dialog name.

courtrightd:
Scripts to fix for this:

- BILS updater: complete
- NOTES - Vendor, currently testing solutions
- NOTICES - LTC Asset Transfer, currently testing solutions
- LTC - Intake Approval: Will be retired with #1517 
- LTC - COLA Summary: Retired
- LTC - MA Approval: Retired

courtrightd:
Notes - Vendor had other problems, moved to its own issue. 
All scripts effected are now updated and ready for testing, or on deck for retirement.
This issue is resolved when #1549 and #1233 are completed.

PR #1551, opened 2023-10-27
#1549 NOTES - Vendor fixing dialog looping

courtrightd:
When searching for a vendor, the script now goes back to the main dialog, allowing additional vendors to be searched. Also allows the script to continue correctly if timed out of maxis, and works with updated check_for_maxis for #1166


C-Love:
I know I'm not actually tagged to review, but I gave this a quick look/test and it looks good to me.

Issue #1554, opened 2023-10-31
Bug in HC evaluation retro request


courtrightd:
In certain instances the script errors when pushing the retro button, giving a subscript out of change. A variable is named incorrectly, so the script usually functions fine because the value stored in that variable works. Sometimes errors. Fix identified.

PR #1555, opened 2023-10-31
#1554 bug in HC evaluation

courtrightd:


MeganGeissler (approved):
Tested and it resolved the error at Retro.

PR #1556, opened 2023-10-31
#1554 fixing problems with the IMIG portion of the script

courtrightd:
Left this under same issue, IMIG was having trouble with misaligned hh member arrays - added a loop to check and match member numbers before evaluating the values in the HEALTH_CARE_MEMBER array.


IlseFerris (approved):
Tested and doesn't appear to break. I approve thee! I'll pull in and add changelog.

Issue #1561, opened 2023-11-06
NOTES - SMRT fix typo for writing memb_numb on STAT/MEMB


MDR22:
There's a typo on line 96 which results in the age of the second person in the HH being written to the CASE/NOTE rather than the selected HH member. Should be writing to 20, 76 not 27, 76.

PR #1562, opened 2023-11-06
#1561 Fix typo in write_value_and_transmit on STAT/MEMB

MDR22:
Fixed typo where write_value_and_transmit was writing to wrong location on STAT/MEMB which resulted in the wrong age (the next HH member's) being written to CASE/NOTE.


courtrightd (approved):
Reviewed, tested, looks good!

Issue #1566, opened 2023-11-07
ACTIONS - TRANSFER CASE: bug in excluded time HC cases


IlseFerris:
If a case is MA Excluded time, that field needs to be completed vs. Health Care CFTR and new Financial Responsibility date.

IlseFerris:
No issues reported, closing.

PR #1567, opened 2023-11-07
#1566 ACTIONS - TRANSFER CASE: Fixed MA excluded time bug

IlseFerris:
Excluded time cases for MA need to be completed on line 15, and this handling was not in the IF...statement.


courtrightd (approved):
Reviewed and tested, works great.
I humbly submit my approval, your highness.

PR #1570, opened 2023-11-14
1569 caf addendum bug

IlseFerris:
I have a live case to test if needed.


courtrightd (approved):
Reviewed and tested looks good

Issue #1572, opened 2023-11-15
Ex Parte Report - Set up Excel for Restarts


C-Love:
There are some Excel files the script creates. Update the Ex Parte Report to read if the files exists and have the script save them when created and updated. This way the reports are not so manual to deal with. 

This is an update in Prep 1 and Prep 2 (I think).

C-Love:
Additionally, add the emailing here

C-Love:
This is working.

PR #1576, opened 2023-11-20
#1572 ADMIN - Ex Parte - set up verifs list to save first

C-Love:
Updating the script to do the following things:

1. Create the Excel files at the beginning of the run and save it. 
2. If the script is restarted, check to see if the Excel File has been already created and reopen.
3. Update the spreadsheets as the script runs. 
4. Once the list is fully processed, save and close the files. Then email these files as needed.

This is basically a restructure of how the Excel is handled because we want it handled more programmatically.


IlseFerris (approved):
No other questions about this one. I approve thee!

Issue #1553, opened 2023-10-31
ACTIONS - PF11 ACTIONS enhancement, needs to be in prod


IlseFerris:
@MDR22 and I were sending a PF11 and discovered that you cannot send them in inquiry mode. Need to add handling to ensure are in production mode.

PR #1578, opened 2023-11-20
#1553 Ensure script is only run in production

MDR22:
Added handling to navigate to SELF and then check MAXIS environment. Script will end if not run in production.


IlseFerris (approved):
Tested and working, I approve thee!

PR #1579, opened 2023-11-20
1446 remove ltc button from notes main menu and include LTC scripts in alpha categories

MeganGeissler:
Removed LTC button from Notes Main Menu
Updated ChangeLog to notify users of this change.
No action required for CLOS or ? for sharepoint documents.
Did not rename, left LTC in the naming for now.


IlseFerris (changes requested):
Hey Megan, I should have mentioned this before but there's some handling in the COMPLETE LIST OF SCRIPTS that should also get updated. Instructional links might be the only thing, but check with @courtrightd as well.

MeganGeissler:
@IlseFerris I updated CLOS but check it out in case there is a better way to update this.

MeganGeissler:
@IlseFerris I reverted CLOS so I can merge this separately after I pull a fresh copy of the CLOS

IlseFerris (approved):
I approve thee!

Issue #1446, opened 2023-08-17
NOTES - MAIN MENU: Update to include LTC into regular alpha structure


IlseFerris:
Currently we have 2 separate buttons for LTC and IMIG based scripts in the NOTES - MAIN MENU, See #1234. With several scripts being incorporated into single script functionality and upcoming retirements, it's time to update these. Tasks will include:

- [x]  Updating the Complete List of scripts
- [x]  Reviewing/updated the automated functionality that creates the buttons. No action required.
- [x]  Update instructions on SharePoint so ?'s still work in the NOTES - MAIN MENU dialog.
- [x]  Create changelog communicating change to staff.

MeganGeissler:
Removed LTC button from Notes Main Menu
Updated ChangeLog to notify users of this change. 
No action required for CLOS or ? for sharepoint.

PR #1582, opened 2023-11-20
#1446 Updated CLOS for LTC

MeganGeissler:
Updated CLOS for LTC by removing specific handling for LTC since documents are now stored in NOTES folder with all other notes scripts


IlseFerris (approved):
All the instructional links are working: I approve thee!

Issue #1540, opened 2023-10-18
Actions - ME-EPD Fiater phrasing improvements


courtrightd:
On the income dialog, the script currently says "No UNEA found" if there is not unemployment income on the case. The wording should be updated to be more clear that it is specific to unemployment income, these case often have other unearned income such as RSDI, which is not clear with the current wording.

PR #1585, opened 2023-11-20
#1540 fix wording and allow full sentence to display.

courtrightd:
Changed the script to say Unemployment income instead of UNEA.
Changed the text field length so that it shows the full statement "The script found the following unemployment income:" Previously only showed "The script found the following"


IlseFerris (approved):
I approve thee!

PR #1597, opened 2023-12-04
1594 december annual updates

IlseFerris:
Ten day cut off dates are listed in TE19.132


MDR22 (approved):
No changes - looks good!

MeganGeissler (commented):
I was doing manual math to verify that TE made sense and Nov 24 doesn't make sense to me. For Nov 24: Day 10 = 11/21 therefore batch would be on 11/20. Could be due to capitation date but I couldn't find this on SIR to verify. Only Nov-Dec of 23 and Jan of 24 calendars are available. 

All others made sense (assuming capitation wasn't a factor).

IlseFerris:
Hey @MeganGeissler here's what TE19.132 says about how the dates are figured out (you totally had it with capitation): 

> 10-day cutoff is derived by counting back 10 days, starting with the
> last day of the month, including holidays and weekends.  Batch must take
> take place prior to the 10-day cutoff allowing for printing, inserting,
> sorting and mailing of the notice.  The postmark on the envelope must be
> 10 days before the end of the month.
> 
> If 10-day cutoff falls on the same day as the PPHP Capitation date 10-
> 10-day cutoff is moved to the day before.

Issue #1601, opened 2023-12-07
NOTES - HC Eval - bug if 'complete application' was pushed to early


C-Love:
Report from AGK:

Script name - NOTES - HEALTH CARE EVALUATION.vbs was run on Case #2616541 with a runtime of 217.2891 seconds. 
Information: accidentally hit the complete evaluation button too early, then selected "no" to go back to the script and after that the "complete evaluation" confirmation dialog popped up every time I switched between different pages of the script

C-Love:
This is done, no issues reported.

PR #1605, opened 2023-12-15
#1601 NOTES - HC Eval leave loop for complete - boolean update

C-Love:
Bug reported by AGK that if 'Complete' was pressed, then 'No' - complete would come up with every button push - which is annoying. This was a boolean reset issue.


IlseFerris (approved):
Looks good. I approve thee!

Issue #1168, opened 2023-02-16
NOTES - Interview needs a check_for_MAXIS after dialogs once function dialog is updated


C-Love:
The check_for_MAXIS updates being completed in Issue #1166 will allow check_for_MAXIS to be added to NOTES - Interview after the dialogs to be sure we are still in the MAXIS system as the script continues.

IlseFerris:
Work on this after #1166 is completed.

C-Love:
This appears resolved with no isuses.

PR #1608, opened 2023-12-15
#1168 NOTES - Interview add check_for_MAXIS

C-Love:
We want to make sure someone hasn't navigated out of MAXIS before running the next part of the script. 

NOTE: there is only 1 call after the large series of dialogs that happens for the main dialog AND for the verbal review of required interview information. 

ALSO NOTE: there is not one after the check eDRS because eDRS does not appear to be in MAXIS and I want the dialog to stop at this screen


IlseFerris (approved):
Looks good. I approve thee!

Issue #1602, opened 2023-12-07
MA FiATer for GRH/MSA - does not correctly deal with COLA


C-Love:
Error reported by TE:
getting a script error related to COLA when running the MSA GRH FIATer. It then aborts the script so I am not able to send a report via MAXIS. Even though it aborts, it has still changed the budget from an X budget to a B budget. Here is a screen shot of the error and where it is in the process. (see team email)

Need review of the bug.

PR #1610, opened 2023-12-15
#1602 - ACTIONS - MA FIATer for MSA/GRH - error with COLA and SSI 

C-Love:
The script was trying to FIAT SSI income and include a COLA deduction - which isn't needed for SSI income since it is excluded entirely. 

Added handling to not do the COLA thing if it is SSI.


C-Love:
Had to redo the logic because TE - who reported the error had a good point that I didn't accommodate for initially.

IlseFerris (approved):
Looks good, I approve thee!

Issue #1282, opened 2023-04-29
Interview Waiver Evaluation for Script Project Impact


IlseFerris:
On April 26, 2023, the Food and Nutrition Service (FNS) approved Minnesota's request to waive SNAP application and/or recertification interviews. The purpose of the waiver is to provide administrative relief as healthcare renewals begin again and offer counties and Tribal Nations more flexibility to meet client needs. The waiver is approved from May 1, 2023 through April 30, 2024 – a full year.

Several issues and initiates should be in place to make this happen including: 
- Making the interview script mandatory
- Add supports to the NOTES - CAF script to CASE/NOTE interviews waived for application and renewal. 
- prioritizing several issues

•	[#719](https://github.com/Hennepin-County/MAXIS-scripts/issues/719) Create note to document attempted contact that are NOT interviews. 
•	[#1190](https://github.com/Hennepin-County/MAXIS-scripts/issues/1190) - Update Client Contact to prevent use of interview documentation
•	[#560](https://github.com/Hennepin-County/MAXIS-scripts/issues/560) - Interview to handle different application forms

C-Love:
We need to figure out how all of this will be evaluated. 

One thing we are going to do right away is change the Appointment Notice functionality to only send the notices on MFIP cases and send renewal guidance on all others.

We will complete this starting with 07/23 Renewals.

PR #1616, opened 2024-01-02
#1282 New Script - SNAP Waived Interview

courtrightd:
Pull request has the new script and CLOS updates.


IlseFerris (approved):
I didn't test the new script, but I think we're good. I approve thee!

PR #1620, opened 2024-01-04
#1282 snap waived interview bug fixes

courtrightd:
Fixes numerous bugs in dialogs and case noting.


C-Love (approved):
Two things:
1. On the person dialog, the imig status box does not align with the header above it. 
2. The 'add verification' for the job question didn't work (Q9 on the CAF I think ) - I think this is because I didn't add any specific jobs.

Issue #1624, opened 2024-01-05
BUG in MA FIATer for MSA GRH if there are months missing in the budget span


C-Love:
There is a bug that is coming up if there are months in the beginning of the budget span because the script thinks that it should find a date when there is no date. 

Need to add condition to the script when reading the date variable to ensure we are still seeing a date.

C-Love:
This is resolved, no additional errors reported at this time. Closing.

PR #1625, opened 2024-01-05
#1624 ACTIONS - MA FIATer for MSA GRH - bug when reading date

C-Love:
This area of the script loops through the budgets of 6 months that were already marked, it assumes there are always 6 months to review but if there are fewer than 6, the script errors when it reads information that should be a date and is not a date. It also transmits too far if the moving is not conditional as well. 

This resolves team email from TE on 1/4/24. 

I have selected everyone for review, but I just need a couple people to give it a try. The case in the team email can be put through background and this script tested without long term impact on the case.


C-Love:
With the approval of the chickens, I am going ahead.

Issue #123, opened 2020-07-26
Create Eligibility Summary Script


C-Love:
Needs a script that combines Approved Programs, Denied Programs, Closed Programs. 

This should include more detail about HOW the budget was determined. 

This may be an 'APPROVER' script.

Once this is done we need to change NOTES - HRF to Processing HRF Form or something along those lines because the APPROVAL of a HRF will be noted with Eligibility Summary. 

We will also need to remove NOTES - DWP Budget because this information will be in Eligibility Summary

C-Love:
Look very strongly at making this an APPROVER.

C-Love:
Moving this to LATE Q4 - possibly Q1 depending on other projects.

C-Love:
Include information about screening for MA EPD when DENYING for HC

C-Love:
Include note specifics about MA EPD Med B manual reimbursement requests and determinations. This should enter in CNote and PNote.

C-Love:
SEND MEMO to Clt at APP with  Reporting Detail based on the APP

C-Love:
Make sure to include any cash Opt Outs and read PACT for that relevant information in determining the budget to case note.

C-Love:
Add review of TE 02.05.108 - Denying/Closing SNAP for No Income Verification

This outlines that the PIC must be $0 income when denying or closing for no verification. This creates a correct notice to the resident about why we are denying. We need Eligibility Summary to check the PIC about income when denying or closing snap to make sure we have coded things correctly.

C-Love:
Check SNAP cases for CCAP and ensure cat elig/DCEX is correct. https://hennepin.sharepoint.com/teams/hs-economic-supports-hub/SitePages/Episode-43.aspx

C-Love:
Include GRH/HS specific handling.

IlseFerris:
Support for aligning non-HC programs to renew together. Identify 24 month certification SNAP folks.

C-Love:
Check ACCT and CASH to ensure assets are counted for SNAP.

IlseFerris:
Consideration for SNAP - Setting TIKLs or reminders automatically for anyone that is a 24 month reporter.

IlseFerris:
Another consideration for SNAP - MA-EPD premium/spenddown/waiver obligation/Medicare costs as FMED deductions.

C-Love:
Include MFIP Special DIET MONY/CHCK approval.
[...]

PR #1627, opened 2024-01-05
#123 NOTES - Elig Summ - verif less than 10 days for Fail to file

C-Love:
Error reported on ELig Summ for times when a verification request can be sent the same day a case is processed - when the case is also ineligible for fail to file since notices have already been sent. 

Script now allows this 10 day requirement to be skipped when HRF or REVW is incomplete


IlseFerris (approved):
Looks good. I approve thee!

PR #1631, opened 2024-01-10
#1282 snap waived interview bug fixes and enhancements

courtrightd:


IlseFerris (approved):
Looks good. I didn't run it but gave it a view. Just throw a change log in there and we're all set. I approve thee!

PR #1635, opened 2024-01-16
1500 eib attach file to error emails

MeganGeissler:
Updated 2 files: Master Func lib and Report to BZST with functionality to attach txt file to error emails for script end procedure w/ error reporting and report to BZST.


IlseFerris (approved):
@MeganGeissler you and I went over this verbally, but I added the Boolean statements and added the option to select the script directly. It's working as intended now. I approve thee!

Issue #1622, opened 2024-01-04
SNAP Waived Interview - Create Return Contact Functionality 


C-Love:
With the creation and implementation of the NOTES - SNAP Waived Interview script in #1282. We have a need for pulling the missing information during a contact after the initial attempt. 

This script will read CASE/NOTE to find the missing information and present it in a dialog for review. This can be used if the client calls back to provide additional details to support SNAP processing. 

This functionality will likely be added to the NOTES - SNAP Waived Interview, but we will create access to it from NOTES - Client Contact when #1621 is completed.

PR #1637, opened 2024-01-19
#1622 Add Return Contact to SNAP Waived Interview

C-Love:
Added functionality to read the CASE/NOTE from the SNAP Waived interview script run to pull the details back into a dialog so that we can complete the contact details when we do get a hold of the resident (on a future call or return call from the resident).


courtrightd (approved):
I have tested this a fair bit this morning, seems to be working well.

Issue #1633, opened 2024-01-10
BUG on NOTES - Application Received if a 'single quote' mark is in case information


C-Love:
SQL will error if any string includes a single quotation mark or apostraphe because this indicates the end of the string in SQL. 

Error if the case name includes a single quote. Need to use a replace this character with another character or something else.

C-Love:
This change is in place and didn't seem to break anything else. Testing for this is unlikely to come back with any detail.

PR #1640, opened 2024-01-22
#1633 NOTES - App Received - replace single quote for SQL

C-Love:
Added a 'replace' to remove any apostrophe in the case name and replace with a dash. This is required for the entry into SQL to save pending case information and potential expedited information for assignments. 

SQL does not allow for single quotes/apostrophes in the strings.


IlseFerris (approved):
Looks good, I approve thee!

Issue #1643, opened 2024-01-23
BUG in HC Eval for date handling


C-Love:
There have been some error reports in HC Evaluation. 

The report had a cdate error but it has been unclear as to how it was triggered. 

Finally found that the baseline date might be entered as a non-date and this might cause this error if it is not blank but is defined as something other than blank. Also discovered that there is a variable mismatch in the application_date variable used in the LTC application functionality around dates. 

Need error handling in the main dialog to ensure that baseline_date is defined as a date if it is not blank. 

Additionally need to align the application_date variable to hc_application_date.

C-Love:
This appears to be resolved.

PR #1644, opened 2024-01-23
#1643 NOTES - HC Eval - BUG in baseline date if not a date

C-Love:
Error had been reported with a cdate error. I believe this is due to handling around the baseline date, particularly needing dialog handling to ensure this is actually a date if it isn't blank. 

Also found an issue with the application date variable.

I selected everyone for review but I think having 2 reviews of the pull would be sufficient.


IlseFerris (approved):
Looks like it should work. I approve thee!

courtrightd:
Looks good from my house! 
I tossed in a typo fix on the dialogs while I was in there.
Approved.

Issue #1646, opened 2024-01-23
MA FIATer for MSA/GRH buggy handling for Deemed Income


C-Love:
Report received that the HH Member Selection dialog when selecting the deeming member does not operate reliably. The persons have to be checked/unchecked repeatedly until the selection is made. 

This dialog needs to be pulled from the FuncLib function and handled locally so that we can use the Dialog1 conventions. This will also allow us to make the dialog more specific to the deeming member selection. 

In reviewing this report, it also seems that the dialog for displaying the deemed income does not always display in the best way.

C-Love:
This appears fixed, we will close this and if any errors are reported, we can reopen the issue.

PR #1647, opened 2024-01-23
#1646 ACTIONS - MA FIATer for MSA-GRH - deeming functionality minor bugs

C-Love:
Changed the dialog parameters for deeming income and updated the dialog used for the deeming members selection.

Additionally moved the testing_run to be true ONLY when deeming income was selected. We seem to have sufficient testing for the standard run. 

If you need a case to test on the error report from TE in the team email has one with 2 members that deeming functionality can be tested. 

Selected all team members for review, but I am looking for two reviews.

### DIFF
--- a/actions/ma-fiater-for-grh-msa.vbs
+++ b/actions/ma-fiater-for-grh-msa.vbs
@@ -751,6 +751,7 @@ Dim client_array
 
 DO
 	' Getting the individual on the case
+	client_array = ""
 	CALL HH_member_custom_dialog(HH_member_array)
 	IF ubound(HH_member_array) <> 0 THEN MsgBox "Please pick one and only one person for this."
 LOOP UNTIL ubound(HH_member_array) = 0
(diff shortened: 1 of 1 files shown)


IlseFerris (changes requested):
I selected two members, and got the message "only select one member", but then additional checkboxes for MEMB 01 and MEMB 02 got added to the HH_member_dialog, for a total of 4. Did it again and got 6 checkboxes, etc.

actions/ma-fiater-for-grh-msa.vbs
```diff
@@ -747,6 +747,7 @@ Do
 Loop until are_we_passworded_out = FALSE														' }
 
 Call back_to_SELF
+Dim client_array
 
 DO
```
MDR22, line 752: Although you didn't change this, I kept running into a bug where if more than one person is selected, you get a msgbox about selecting one person only. When HH_member_custom_dialog(HH_member_array) is called again, it then adds the HH Membs to the dialog again (so if you have 2 HH members in the dialog to start, you will have 4 after the first msgbox, and then 6 again after it calls HH_member_custom_dialog(HH_member_array) again.

MDR22 (commented):
I tried on the case from the team email and didn't run into any issues besides the one comment related to selecting too many HH members. Looks good!

C-Love:
Ok - initial dialog updated - well blanking out of the client_array before calling the function to select the member to make sure we don't multiply all the hh members. Second review needed.

IlseFerris (approved):
I tested this on the case in the TE, and it worked as expected. I approve thee!

MDR22 (approved):
No issues with duplicating HH Membs now.

PR #1648, opened 2024-01-23
1645 case transfer updates

IlseFerris:


actions/transfer-case.vbs
```diff
@@ -135,24 +136,36 @@ DO
...
+    EMReadScreen transfer_message, 79, 24, 2
+    transfer_message = trim(transfer_message)
 
     STATS_manualtime = 120                      'manual run time in seconds
-    'Script low down for innercounty transfer
+    'Script low down for inner county transfer
     script_run_lowdown = script_run_lowdown & "transfer_out_of_county: " & transfer_out_of_county & vbCr & "worker_number: " & worker_number & vbCr & "transfer_to_worker: " & transfer_to_worker & vbCr & " Error Message at transfer: " & vbCr & error_message
     If servicing_worker <> transfer_to_worker THEN script_end_procedure_with_error_report("Transfer of this case to " & transfer_to_worker & " has failed.")
-    script_end_procedure_with_error_report("Success! Case transfer has been completed to: " & transfer_to_worker & ".")
+    script_end_procedure_with_error_report("Success! " & transfer_message)
```
C-Love, line 168: I think this might be an unclear message on it's own for the users. Helpful for us in reviewing any error reports, but it probably needs some context in the end message.

COMPLETE LIST OF SCRIPTS.vbs
```diff
@@ -2249,6 +2249,7 @@ script_array(script_num).tags                   = array("BZ", "Monthly Tasks", "
 script_array(script_num).dlg_keys               = array("Up", "Ex", "Ev")
 script_array(script_num).subcategory            = array("")
 script_array(script_num).release_date           = #10/01/2000#
+script_array(script_num).retirement_date        = #01/23/2024#
```
C-Love, line 2252: Are we moving inactive transfer to the retired scripts?

C-Love (changes requested):
Some housekeeping and context handling.

actions/transfer-case.vbs
IlseFerris: The message does have some decent information. Here's an example from a training case: CASE XFER'D FROM X127GM1 TO X127L1S ON 01/23/24 AT 13:39:04

COMPLETE LIST OF SCRIPTS.vbs
IlseFerris: The BULK - INACTIVE TRANSFER is being retired temporarily while the 'excluded' baskets and the entire need of transferring cases is re-evaluated.

IlseFerris: Nope, this will take a minute I'm sure.

actions/transfer-case.vbs
C-Love: I'm thinking something more like this:
image

This would also be great for if it failed as well. 

I just know that I have struggled getting users to understand when information is coming from MAXIS. I have a commit with some updates.

Issue #1645, opened 2024-01-23
Case Transfer Process Supports 


IlseFerris:
Met with product owners for ES Workflow and ECF Next today. There's a known issue with order of operations adherence with users when transferring cases cases in MAXIS. Staff are supposed to transfer cases 1st in ECF Next then in MAXIS to ensure that document notifications are still attached to the work item. Staff that are not adhering to this process are causing issues in work not being visible or known to the workers using the case load structures in ECF Next or through ES Workflow. 

Will be making the following updates to support adherence and reduce errors in other associated processes: 
- [x] BULK - CASE TRANSFER: Moving to restricted ADMIN menu. Based on current instruction and review of usage, only certain OS staff should be using this script at this time. 
- [x] ACTIONS - TRANSFER CASE:  
    1. Add confirmation of transferring case in ECF Next prior to continuing to the case transfer. 
    2.  Update closing message to include case load transferred from as well. 
- [x] ADMIN - INACTIVE TRANSFER: retire until inactive caseloads and associated ECF Next notifications are figured out.

IlseFerris:
No issues reported on updates. Closing issue.

PR #1649, opened 2024-01-23
#1645 ACTIONS - Transfer Case - closing message additional verbiage

C-Love:
Adding come additional verbiage for the inner county closing message.


IlseFerris (approved):
This looks great, thank you and I approve thee!

Issue #1650, opened 2024-01-23
Record Script Usage outside of Script End Procedure function


C-Love:
Currently, the only way we have to record if a script is run is if we get to the end of a script run and the SQL INSERT to the dbo usage log is triggered. There is no reason we can't enter records outside of the script_end_procedure and this would be made easier with a new function meant to record script usage. 

First, we are going to create this function and test it in the updates to the DAIL scrubber with COLA Review and Approve. 

Once we confirm that this works on it's own, we can change both script_end_procedure functionalities to call this function to be sure it works to in this instance as well. 

We will have to make sure the reporting doesn't change with this new functionality.

C-Love:
Testing seems complete on this, no issues reported and the functionality is in play in Client Contact. 

Leaving this in discuss at next meeting to review the functionality naming conventions and structure of recording the parts of functionality. Need a plan to make this cohesive.

C-Love:
DaNM - this functionality generally works but our naming and verbiage conventions are not created. Do we want to close this issue since the functionality is done? Or should we spend a little time finishing these conventions?

MeganGeissler:
> DaNM - this functionality generally works but our naming and verbiage conventions are not created. Do we want to close this issue since the functionality is done? Or should we spend a little time finishing these conventions?

The only insight I've received from the PowerBI team regarding where dbos.usage.log file is that it is used as the underlying data file for the data sources: HsScriptFact and HsScriptDim which are the tables they use to build our power bi graphics. 

Beyond that, we don't have much insight into where else this data file is being used and how changing the naming conventions would impact other areas. Considering this, we could wrap this up by deciding on a naming convention for function/incremental script measurements and provide these categories to Ernie so he can incorporate it into our powerbi dashboard. Thoughts and hashing out naming format and wrapping this up?

MeganGeissler:
Spoke with ME- You cannot see from the underlying data source (such as dbos.usage.log) where it is being used. You have to work backwards from a dashboard. Processing dashboard, activity report, and BZ report are the sources ME is aware of that use dbos.usage.log. 

If we add something at the beginning of our naming convention such as "FUNCTION" this would be easy for them to exclude or group in a different way in order to display it in say our BZ report graphics, however tacking additional words onto the end (DAIL -COLA REVEIW AND APROVE - COLA SUMMARY NOTE CREATED) would be trickier to handle for.

C-Love:
[...]

Issue #1518, opened 2023-10-09
DAIL - COLA Review and Approve - update to create an income summary


C-Love:
Change the functionality of DAIL - COLA Review and Approve so that it will create an income summary in CASE/NOTE for ease of review. 

We will be removing all functionality associated with creating a CASE/NOTE of an approval as that will be supported by Eligibility Summary. 

The current COLA Income Summary is in the script LTC - COLA Summary. This script also has approval processing but that will not be maintained in the DAIL scrubber. 

This is connected to #1233 and removing all other forms of MAXIS APP functionality supports.

C-Love:
This appears to be functioning. Additional output data is being used on the pieces of functionality.

PR #1651, opened 2024-01-23
#1650 and #1518 - New usage recording functionality

C-Love:
Created new function to record usage information outside of the script_end_procedure so that we can record more specific script functionality usage. 

This is an initial phase of this work and we will need to consider how we use the table columns that currently exist.

We might need to change the naming parameters when saving a part of a script run, like add 'FUNCTIONALITY - ' to the beginning or something - or maybe remove the '.vbs' and add something else. 

I also have some questions about resetting the timer. We want to be careful we are not double counting script run time. Currently I have a boolean parameter that will allow us to control resetting the timer.

Here is what the data looks like with my recording the usage in this way:
image
The last two are from the same script run.

We should probably talk about this in more detail but any update should have minimal impact on scripts other than COLA Review and Approve so we could pull in and discuss later, especially if we need the FuncLib available.

### DIFF
--- a/MASTER FUNCTIONS LIBRARY.vbs
+++ b/MASTER FUNCTIONS LIBRARY.vbs
@@ -4962,6 +4962,49 @@ function clear_line_of_text(row, start_column)
   EMWaitReady 0, 0
 end function
 
+function collect_script_usage_data(script_to_record, closing_message, reset_timer)
+'--- This function records script runs into the usage log and can be used for functionality runs, not just completed scripts.
+'~~~~~ script_to_record: this would be the 'name_of_script' for completed script runs, but can be adjusted if run on functionality instead of the completed script
+'~~~~~ closing_message: information about the script run should be passed through this variable, for completed script runs, this would be the closing message
+'===== Keywords: MAXIS, MMIS, script, statistics
+	stop_time = timer
+	script_run_end_time = time
+	script_run_end_date = date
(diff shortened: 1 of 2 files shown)


MASTER FUNCTIONS LIBRARY.vbs
```diff
@@ -4962,6 +4962,49 @@ function clear_line_of_text(row, start_column)
   EMWaitReady 0, 0
 end function
 
+function collect_script_usage_data(script_to_record, closing_message, reset_timer)
+'--- This function records script runs into the usage log and can be used for functionality runs, not just completed scripts.
+'~~~~~ script_to_record: this would be the 'name_of_script' for completed script runs, but can be adjusted if run on functionality instead of the completed script
+'~~~~~ closing_message: information about the script run should be passed through this variable, for completed script runs, this would be the closing message
+'===== Keywords: MAXIS, MMIS, script, statistics
```
IlseFerris, line 4969: Line 4969: I'm assuming reset_timer is a Boolean, but can you add instructions for this parameter?

C-Love: I left it out because I'm not sure we need it - or if it makes sense to have it here at all.

IlseFerris (approved):
Just the one little update, otherwise this looks EPIC! I approve thee!

PR #1656, opened 2024-01-25
1606 tlr enhancements

IlseFerris:
UAT updates added for selecting the month/year and updating through CM + 1.

### DIFF
--- a/actions/tlr-screening.vbs
+++ b/actions/tlr-screening.vbs
@@ -54,21 +54,23 @@ changelog_display
 EMConnect ""
 Call check_for_MAXIS(False)
 Call MAXIS_case_number_finder(MAXIS_case_number)
-MAXIS_footer_month = CM_mo
-MAXIS_footer_year = CM_yr
+Call MAXIS_footer_finder(MAXIS_footer_month, MAXIS_footer_year)
 member_number = "01"
 
 Dialog1 = "" 'Blanking out previous dialog detail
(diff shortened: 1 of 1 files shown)


actions/tlr-screening.vbs
```diff
@@ -510,7 +513,7 @@ If exempt_reasons > 0 then
       EditBox 70, 90, 250, 15, exemption_notes
       Text 10, 115, 60, 10, "Exemption basis:"
       ComboBox 70, 110, 100, 15, "Select OR Type..."+chr(9)+"Conversation w/ resident"+chr(9)+"Observational"+chr(9)+"Verified", exemption_basis
-      CheckBox 10, 130, 205, 10, "Check here to update STAT/WREG with highest exemption.", update_wreg_checkbox
+      CheckBox 10, 130, 245, 10, "Update STAT/WREG with highest exemption for selected month - CM +1.", update_wreg_checkbox
```
C-Love, line 516: Just a reminder, the 'Check Here' is preferred because it supports screen readers knowing how to interact with the element.

C-Love (approved):
I did not run this on a case but this appears solid updates. Looks like supportive updates. 

I have one comment on dialog verbiage., but that is completely up to you on if it is important in this scenario.

MDR22 (approved):
Didn't run into any issues while testing in training - looks good!

PR #1661, opened 2024-01-26
1434 tb dail capture updates

IlseFerris:
Cleaned up a lot of the unnecessary code, and added new Closing Issue Documentation. Can be run to test with the default settings for the dialog in place. It won't mess anything up if it's run for assignments more than once.


courtrightd (approved):
Tested fine, and I didn't spot any issues in the code. I approve!

Issue #1667, opened 2024-02-02
UTILITIES- VA Verification Request


MeganGeissler:
Email for VA Verif Request is pulling in the last member on the case's SSN instead of the vet's ssn.

Email from reporter: 
I submitted a VA Verification Request found under the UTILITIES SCRIPT. The request was for M01 and selected this member however when the email was sent it sent with M01 name and M03 SSN. I tried a second time, again when the email was sent with request, it displays M01s name and M03s SSN instead of M01s SSN

PR #1668, opened 2024-02-02
#1667 Replaced client_ssn with vet_ssn for email

MeganGeissler:
Email was using the client_ssn. Replaced client_ssn with vet_ssn in email to align with the name beign pulled in (vet_member) so that the SSN is correct regardless of what member is selected.


MDR22 (approved):
Makes sense - looks good!

IlseFerris (approved):
Tested and worked as intended. Throw in a changelog and we're good to go. I approve thee!

PR #1676, opened 2024-02-09
#1650 NOTES - Client Contact - record redirect to SNAP Waived Intvw in usage log

C-Love:
Adding the output to the usage log when we redirect form client contact to SNAP waived interview. The entries would look something like this:

image


IlseFerris (approved):
Looks good. I approve thee!

Issue #1677, opened 2024-02-14
Ex Parte Bug on COLAs and STILL DUPLICATE UNEA panels


C-Love:
Report that the Ex parte BULK runs do not correctly update UNEA. 

More tweaks, let's see if it works.

C-Love:
No errors have been reported at this time, appears these fixes work better. Closing with no reports.

PR #1679, opened 2024-02-14
#1677 ADMIN - ExParte Report - COLA handling

C-Love:
How about we deal with COLA


IlseFerris (approved):
This looks sound. I approve thee!

Issue #1678, opened 2024-02-14
Determine program and case status function thinks pending EGA is DWP


C-Love:
The function to determine what programs are active or pending has the DWP variable associated with reading EGA instead of DWP. 

This should be resolved with the replacement of this variable.

C-Love:
This appears to be fixed.

PR #1680, opened 2024-02-14
#1678 Functions Library - correct variable for DWP case status

C-Love:
Why did this take me so long to find... ugh - clearly still getting better


MDR22 (approved):
Looks good and makes sense.

IlseFerris (approved):
I approve thee!

Issue #1688, opened 2024-02-22
ADMIN - Resolve HC EOMC - Update phrasing 


C-Love:
There were some questions about the information provided on the 'Budget Error' part of the worklist report. The functionality appears to be working correctly, but we can update the phrasing to provide better information to the workers. 

This issue is for a verbiage update on the generated Excel worklist only.

C-Love:
Update worked. Single user - me- closing.

PR #1689, opened 2024-02-22
#1688 ADMIN - Resolve HC EOMC - Budget Error Worklist verbiage update

C-Love:
Changing some words to make the direction for the work list have a different perspective.


IlseFerris (approved):
Verbiage changes make sense. I approve thee!

Issue #1691, opened 2024-02-22
Add appointment notice function to NOTES - SUBSEQUENT APPLICATION


IlseFerris:
The `create_appointment_letter_notice_application` function needs to be added to the NOTES - SUBSEQUENT APPLICATION script to complete our work on centralizing our notice information. #1685 is associated, and will also be using the functions vs. the office information being hard coded into the scripts.

C-Love:
This appears to be working.

PR #1693, opened 2024-02-22
#1691 - NOTES - Subsequent App - MEMO to use Function

C-Love:
Removed the memo wording to use the function that defines the appt notice wording centrally.


IlseFerris (approved):
Looking good. I approve thee!

PR #1699, opened 2024-02-23
1636 notes proof of relationship resolve hh memb issues

MDR22:
Fix issue where script only displays selected HH members in relationship dialog, rather than displaying all relationship connections between all HH members for the selected HH members. Also fixed the functionality to add a new relationship row. However, this seems to cause issues when the dialog gets too large as the Docs Received field starts to default with info - seems to pull in the "MEMB" checkbox.

### DIFF
--- a/notes/proof-of-relationship.vbs
+++ b/notes/proof-of-relationship.vbs
@@ -147,7 +147,6 @@ selected_HH_member_array = "*"
 FOR i = 0 to total_clients
 	IF all_clients_array(i, 0) <> "" THEN 						'creates the final array to be used by other scripts.
 		IF all_clients_array(i, 1) = 1 THEN						'if the person/string has been checked on the dialog then the reference number portion (left 2) will be added to new HH_member_array
-			'msgbox all_clients_
 			selected_HH_member_array = selected_HH_member_array & left(all_clients_array(i, 0), 2) & "*"
 		END IF
 	END IF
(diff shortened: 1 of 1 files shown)


C-Love (changes requested):
The dialog works really well. 
However, it looks like the case note will add all of the information to the CASE/NOTE and isn't recognizing the checkbox for a new verification or already known. 

The second CASE/NOTE in 332914 (training) was created with only selecting MEMB 03.

notes/proof-of-relationship.vbs
```diff
@@ -203,46 +279,74 @@ Next
...
+		Text 295, 10, 50, 10, "Related to:"
+		Text 375, 10, 75, 10, "Document(s) received: "
+		Text 5, (60 + ((selected_hh_memb_relationship_count + add_field_count) * 20)), 80, 10, "Other verifs still needed:"
+		Text 265, (60 + ((selected_hh_memb_relationship_count + add_field_count) * 20)), 45, 10, "Other Notes:"
+		Text 5, (80 + ((selected_hh_memb_relationship_count + add_field_count) * 20)), 60, 10, "Panel(s) updated: "
+		Text 280, (80 + ((selected_hh_memb_relationship_count + add_field_count) * 20)), 60, 10, "Worker signature:"
+		' Text 5, (35 + (array_item * 20)), 575, 10, "  * This last line is available for entry of relationship proof that was not documented in STAT. If the Relationship is left as 'SELECT ONE...' this line will not case note."
+		EndDialog
+
+		msgbox "add_field_count " & add_field_count
```
C-Love, line 346: Testing msgboxes are here

C-Love (approved):
Ok - it's working well - you just need to remove 2 msgboxes around line 346.

Issue #1686, opened 2024-02-20
Update to HC Eval for Ex Parte Phase 1 - EditBox for Mixed Household Info


C-Love:
Ex Parte team has a manual process to deal with the Mixed Household cases. However, this is causing some difficulty understanding the situation because 2 conflicting CASE/NOTEs are being entered. 

If we can add a field on HC Eval for the ability to enter Mixed Household detail in the note we might avoid having an additional CASE/NOTE. 

Potentially, this could email BT.

C-Love:
This appeared to work for an entire month of Phase 1 and is good

PR #1701, opened 2024-02-23
#1686 NOTES - HC Eval - Add Mixed HH Note

C-Love:
Additional support to Ex parte Phase 1 group. 

This will be difficult to test, but it is an isolated user group, we should be able to control the testing starting the 1st of the March. 

I will need to put together testing details next Wed/Thru for the Ex Parte group.


IlseFerris (approved):
Didn't test, but looks like it will work. Throw a changelog in there, and then I approve thee!

Issue #925, opened 2022-07-29
AUDIT - start_a_blank_CASE_NOTE()


MiKaylaH22:
Problem: users are reporting error messaging coming up inappropriately
Possible resolution(s):
1. look at a PF3 option when inside the case note 
2. dialog one should be removed 
3. add cancel option with error reporting 
4.  look at using maxis screen codes - confirm the correct screen before starting
FMCAMAM1 = Case Notes (NOTE)
FMCAMAM2 = Case Notes (NOTE) inside case note

PR #1702, opened 2024-02-26
#925 AUDIT - start_a_blank_CASE_NOTE()

MDR22:
-Removed unnecessary msgbox commented out
-Changed CASE/NOTE edit mode check to evaluate based on the panel name rather than mode and CASE/NOTE panel 
-Add option to cancel script and submit error report

### DIFF
--- a/MASTER FUNCTIONS LIBRARY.vbs
+++ b/MASTER FUNCTIONS LIBRARY.vbs
@@ -12223,7 +12223,12 @@ function start_a_blank_CASE_NOTE()
 		PF9
 		EMReadScreen case_note_open_check, 8, 1, 72
 		If case_note_open_check <> "FMCAMAM2" then
-            BeginDialog Inquiry_Dialog, 0, 0, 241, 195, "CASE NOTE Cannot be Started"
+            EMReadScreen mode_error_check, 1, 20, 9
+			EMReadScreen PW_error_check, 6, 20, 9
+			EMReadScreen error_message_check, 79, 24, 2
+			error_message_check = trim(error_message_check)
+			
+			BeginDialog Inquiry_Dialog, 0, 0, 241, 195, "CASE NOTE Cannot be Started"
 			ButtonGroup ButtonPressed
 				OkButton 185, 110, 50, 15
 				PushButton 180, 175, 50, 15, "Report Error", report_error_button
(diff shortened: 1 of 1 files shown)


MeganGeissler (approved):
Tested and looks good!

MASTER FUNCTIONS LIBRARY.vbs
```diff
@@ -12221,23 +12221,24 @@ function start_a_blank_CASE_NOTE()
...
-            'msgbox "The script can't open a case note. Reasons may include:" & vbnewline & vbnewline & "* You may be in inquiry" & vbnewline & "* You may not have authorization to case note this case (e.g.: out-of-county case)" & vbnewline & vbnewline & "Check MAXIS and/or navigate to CASE/NOTE, and try again. You can press the STOP SCRIPT button on the power pad to stop the script."
-            BeginDialog Inquiry_Dialog, 0, 0, 241, 115, "CASE NOTE Cannot be Started"
-              ButtonGroup ButtonPressed
-                OkButton 185, 95, 50, 15
-              Text 10, 10, 190, 10, "The script can't open a case note. Reasons may include:"
-              Text 20, 25, 80, 10, "* You may be in inquiry"
-              Text 20, 40, 185, 20, "* You may not have authorization to case note this case (e.g.: out-of-county case)"
-              Text 5, 70, 225, 20, "Check MAXIS and/or navigate to CASE/NOTE, and try again. You can press the STOP SCRIPT button on the power pad to stop the script."
-            EndDialog
+		EMReadScreen case_note_open_check, 8, 1, 72
```
C-Love, line 12224: Could we add a read of the mode and the message at the bottom of the screen? Maybe also the PW? Then if it fails and the error report is selected, we can add it to the 'script_run_lowdown' variable so we get additional information in any error report.

C-Love (changes requested):
This all looks great - I only have a suggestion to give us better information if errors are reported in this functionality.

MASTER FUNCTIONS LIBRARY.vbs
MDR22: Good idea. Added script_run_lowdown variable

MASTER FUNCTIONS LIBRARY.vbs
```diff
@@ -12221,23 +12221,34 @@ function start_a_blank_CASE_NOTE()
...
+		If case_note_open_check <> "FMCAMAM2" then
+            EMReadScreen mode_error_check, 1, 20, 9
+			EMReadScreen PW_error_check, 6, 20, 9
```
C-Love, line 12227: I think this should be EMReadScreen PW_error_check, 6, 21, 15

C-Love (changes requested):
Looks good except the coordinates for reading the PW.

C-Love (approved):
Functionality looks good but the coordinates for reading the PW are off. Once that is resolved, you are good to go.

Issue #1600, opened 2023-12-07
DAIL - New Hire NDNH, ensure script navigates back to DAIL if needed and CASE/NOTE aligns with actions taken


MDR22:
Received email from worker regarding DAIL - NEW HIRE NDNH creating an incorrect CASE/NOTE after the worker navigated away from the DAIL - see below:

"I navigated away from DAIL to review a STAT panel, then when I finished the script is said it did not enter to clear the match, but still wrote a case note saying it did. The script should tell you to go back to DAIL or just stop and tell you to start over."

Review validation for DAIL and creation of CASE/NOTE

MDR22:
Per P&I meeting on 12/14, Casey suggested reviewing DAIL - COLA Review and Approve for nav_in_dail function that ensures script returns to DAIL to find matching message.

PR #1704, opened 2024-02-27
#1600 DAIL - New Hire NDNH, ensure script returns to DAIL message

MDR22:
Add handling to ensure that script returns to same DAIL message if user navigates to a different panel during dialog. Updated function nav_in_dail to read full DAIL message to ensure exact match is found in situations where the shortened DAIL message may have duplicates.

### DIFF
--- a/dail/new-hire-ndnh.vbs
+++ b/dail/new-hire-ndnh.vbs
@@ -75,8 +75,58 @@ EMConnect ""
 EMSendKey "T"   'TYPES "T" TO BRING THE SELECTED MESSAGE TO THE TOP
 transmit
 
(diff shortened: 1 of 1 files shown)


courtrightd (commented):
Needs something before writing the "T" at line 92 to make sure its actually in a DAIL message. Emailed details.

MDR22:
Added additional handling to ensure the script could navigate back to HIRE message and move through DAIL

C-Love (approved):
Looks good to me, I didn't try it but I think it's solid. I say go for it.

courtrightd (approved):
Tested, working good!

Issue #1527, opened 2023-10-10
Create a Clean Up script to move/delete xml files for Case Assignment


C-Love:
In order to support testing parsing the xml and have xml available to us for use in script automation, we had IT pull the xml files from the MNIT Mailbox to a network folder. There is no retention or automation on this folder. 

We should provide support in the form of a utility script for Case Assignment that will move the files to an archive folder based on a date indicated by the CA HSS, who will run this script. The script will also delete the xmls entirely after a certain amount of time (3 months).

I do not believe there is any retention requirements for this folder or the files within it.

PR #1712, opened 2024-03-04
#1527 XML file cleanup: CLOS update

courtrightd:


C-Love (approved):
Looks good - tested and the menu with the unique description is good.

Issue #1711, opened 2024-03-04
NOTES-ELIG SUMM - Deemer Income panel read error


MeganGeissler:
Demmer income popup switches between lowercase and uppercase. Need coding to read for both before exiting loop. 

User reported: I have tried running the elig summary script on this case 4 times and it keeps getting stuck.
Case #2398048 starting in 1/2024.

PR #1715, opened 2024-03-08
#123 & #1711 NOTES - Eligibility Summary - HC Supports and Bug Fix

C-Love:
This is a fe3w smaller supportive changes that were discussed with the Ex Parte Team. 

I do actually want to get this in late morning on Monday so the Ex Parte folks have some time to test it out and for me to put any fixes in while they still have cases. 

Bug resolved by @MeganGeissler - MAXIS and Maxis are not the same thing

Adds a date for 1503 
Looks for Remedial Care to count against a spenddown if MA for MEMB in a GRH with NO GRH program eligibility. This can stop the script but it can also be bypassed. Training case 336398 has a member with a person in a GRH facility. You can test with that/change it as needed if you want to try to break it.


courtrightd (approved):
This looks good from my house.

PR #1722, opened 2024-03-12
1709 DAIL - NEW HIRE NDNH, indicate employer name on INFC dialog

MDR22:
Updated INFC dialog to display employer name


IlseFerris (approved):
Looks good. I approve thee!

Issue #1719, opened 2024-03-09
Adding PRODUCTION checks into REPORT TASK scripts


IlseFerris:
Certain processes that are in the ADMIN REPORT TASKS main menu that are run by OS staff require production access vs. inquiry. The following scripts will have prod checks added: 
- [ ] ADMIN  - DAIL 12 MONTH CONTACT
- [ ] ADMIN - DAIL CCD
- [ ] ADMIN - DAIL FMED DEDUCTION

PR #1727, opened 2024-03-13
#1719 Production check function in admin report scripts

courtrightd:


MDR22 (approved):
Tested and works well!

Issue #907, opened 2022-07-18
ELIG Results to Word


C-Love:
Create UTILITY to save details of ELIG to word document with formatting.

PR #1728, opened 2024-03-13
#907 New Scripts - UTILITIES - Elig Results to Word

C-Love:
New script to create a word document of a single Eligibility Result for any program other than HC. This is supportive of processes such as appeals, overpayments or other claims activities.

I will update CLoS in another pull since this branch is quite old.

### DIFF
--- a/utilities/elig-results-to-word.vbs
+++ b/utilities/elig-results-to-word.vbs
@@ -46,6 +46,29 @@ call changelog_update("06/02/2021", "Initial version.", "Casey Love, Hennepin Co
 changelog_display
 'END CHANGELOG BLOCK =======================================================================================================
 
+Function copy_elig_to_array(output_array)
+	output_array = "" 'resetting array
+	For row = 2 to 21
+		EMReadScreen elig_line, 80, row, 1
+		output_array = output_array & elig_line & "UUDDLRLRBA"
+	Next
+	' MsgBox output_array
+	output_array = split(output_array, "UUDDLRLRBA")
+End function
+
+function insert_page_break_after_two_panels(screen_on_page)
+	'Determines if the Word doc needs a new page
+	'screen_on_page - This is a running counter that is updated in this function
+	screen_on_page = screen_on_page + 1
+	If screen_on_page = 3 then												'if we are at 2, we need to insert a page breakk and reset the counter
+		screen_on_page = 1
+		objSelection.InsertBreak(7)
+	ElseIf screen_on_page = 2 Then
+		objSelection.TypeText vbCr & vbCr & vbCr & vbCr & vbCr
+	End if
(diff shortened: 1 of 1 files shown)


courtrightd (commented):
Pushed a commit that fixes an error on GA results, tested good otherwise. 
Might want to adjust the date on changelog initial version.

MeganGeissler:
Ran for ELIG/EMER and it kept rewriting EMSM results. I think you need one more handling for worker_message: that ends with " **"
If worker_message = "** PLEASE PROVIDE A COMMAND OR PF-KEY TO CONTINUE **" Then last_elig_screen = True

ELIG word doc

utilities/elig-results-to-word.vbs
```diff
@@ -0,0 +1,106 @@
...
+		If InStr(MX_line_3, "DWPR") Then elig_results_program_found = "DWP"
+		If InStr(MX_line_3, "MFPR") Then elig_results_program_found = "MFIP"
+		If InStr(MX_line_3, "MSPR") Then elig_results_program_found = "MSA"
+		If InStr(MX_line_3, "GAPR") Then elig_results_program_found = "GA"
+		If InStr(MX_line_3, "CAPR") Then elig_results_program_found = "Cash Denial"
+		If InStr(MX_line_3, "GRPR") Then elig_results_program_found = "GRH"
+		If InStr(MX_line_3, "FCSM") Then elig_results_program_found = "IV-E"
+		If InStr(MX_line_3, "EMPR") Then elig_results_program_found = "EMER"
+		If InStr(MX_line_3, "FSPR") Then elig_results_program_found = "SNAP"
+		If elig_results_program_found = "" Then MsgBox "MAXIS must be at the first page of Eligibility Results for this script to run." & vbCr & vbCr & "The dialog will return." & vbCr & vbCr & "NAVIGATE TO ELIG RESULTS WHILE THE DIALOG IS UP."
```
MDR22, line 118: Not sure if it's necessary but wondered about handling to ensure that the script is pulling results for the case number initially entered. I put in a case number, received the notice to move to elig results, but then navigated to elig results for a different case number. It printed the results for the new case number as expected, but thought about whether an error message of some sort might be useful if the elig results are for a different case number.

MDR22 (approved):
Looks good!

IlseFerris (changes requested):
Good feedback by @MeganGeissler and @MDR22. I don't have any additional comments besides theirs.

C-Love:
I think with all the variations in the last page handling, we needed something more dynamic - so I updated this to using an InStr.

utilities/elig-results-to-word.vbs
C-Love: Good thought - I think with this script we don't really care what the case number is. Removing it entirely from the dialog.

IlseFerris (approved):
Tested on FS and MFIP results and working as intended. I approve thee!

Issue #1717, opened 2024-03-09
Update function determine_200_percent_of_FPG for April 2024 Changes


IlseFerris:
New amounts will go into effect on 04/2024. This has been confirmed in the 04/24 Future CM Changes posted earlier this month.

PR #1734, opened 2024-03-14
#1717 Update function determine_200_percent_of_FPG for April 2024 Changes

MDR22:
-Update 200% FPG for April 2023 for EA and EGA
-Separate programs for easier future 200% FPG updates 
-Update handling for EA and EGA applications with no date to determine 200% FPG amounts based on date of script run
-Update EGA to only handle for applications before or after 4/1/24 (not before 4/1/23)


IlseFerris (approved):
Looks good. I tested this with NOTES - EMERGENCY and got the results I expected. I approve thee! I'll also be pulling this in since I need to update the functions for notices today.

Issue #1744, opened 2024-03-19
SNAP Waived Interview string error when pressing no contact button


courtrightd:

courtrightd:
Re-opening issue, believe I have a correct solution now.

PR #1745, opened 2024-03-20
#1744 Waived Interview no contact 

courtrightd:
This is to resolve a bug that intermittently creates a type mismatch error for certain users when pressing the no contact made button on the information needed dialog.


IlseFerris (approved):
Looks good. I approve thee!

Issue #1634, opened 2024-01-11
NOTES - HRF - add option to indicate NO HRF for next month when processing late


C-Love:
Create an option with a checkbox when the script is being run during the benefit month for the provided HRF. This would add a line to the CASE/NOTE to indicate that the next month's HRF has not yet been received and is being closed because both months have to be processed at the same time. 

Need to confirm the dates for specific processing for the functionality on this process. 

This may also be needed in the CAF script in the future. Additional details and information should be handled in Eligbility Summary.

C-Love:
Appears to be working, no reports of errors.

PR #1756, opened 2024-03-27
#1634 NOTES - HRF - add option for future HRF not received

C-Love:
Creates a line in the CASE/NOTE to indicate that a HRF for a future month is not received when running the script for a HRF for the current month.


IlseFerris (approved):
These updates make sense. I approve thee!

Issue #1250, opened 2023-04-20
NOTES - Eligibility Summary - Warning if a program is 'U' Code for REVW or MONT


C-Love:
This would stop the script run if a REVW or MONT U Code is in place and alert them to complete an approval or change the code.

If status is "U" code in REVW or MONT, we need a notification to the user: The following programs have a review status as "U" (list programs). Complete approval or change status to "i". 

Inquire whether we want to place this in a do/loop to ensure they have fixed the issue in order for the script to proceed.

PR #1760, opened 2024-03-28
#1250-Added handling for "u" status on REVW and MONT panel

MeganGeissler:
Added handling for "u" status but not sure how to embed within the current coding. @C-Love - can you review and give feedback? I left a few questions within the code as well.

### DIFF
--- a/notes/eligibility-summary.vbs
+++ b/notes/eligibility-summary.vbs
@@ -23444,6 +23444,82 @@ class stat_detail
 		stat_mont_form_recvd_date = replace(stat_mont_form_recvd_date, " ", "/")
 		If stat_mont_form_recvd_date = "__/__/__" Then stat_mont_form_recvd_date = ""
 
+		'Herehere
+		'TODO: Is this okay placement? 
+		'TODO: Why does this only work if inside the sub?
+		'TODO: Why does it take so long to initially load the script?
+		'TODO: Do we like how it double checks the value or do we not want to read the statuses twice (once above and once within the error handling). Could restructure:
+			' Do	
+			'	READ Ucodes for REVW and MONT
+			' 	Exit do if U code <> for any of the statuses (this would skip the dialog below) 
+			' 	Do/Loop Dialog & error handling
+			'Exit Do
+
+		If stat_mont_cash_status = "U" Then mont_u_code_string = mont_u_code_string & ", Cash/GRH"
+		If stat_mont_snap_status = "U" Then mont_u_code_string = mont_u_code_string & ", SNAP"
+		If stat_mont_hc_status = "U" Then mont_u_code_string = mont_u_code_string & ", HC"
+		If left(mont_u_code_string, 1) = "," Then mont_u_code_string = right(mont_u_code_string, len(mont_u_code_string)-2)
+
+		If stat_revw_cash_code = "U" Then revw_u_code_string = revw_u_code_string & ", Cash/GRH"
+		If stat_revw_snap_code = "U" Then revw_u_code_string = revw_u_code_string & ", SNAP"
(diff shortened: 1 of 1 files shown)


notes/eligibility-summary.vbs
```diff
@@ -23444,6 +23444,82 @@ class stat_detail
 		stat_mont_form_recvd_date = replace(stat_mont_form_recvd_date, " ", "/")
 		If stat_mont_form_recvd_date = "__/__/__" Then stat_mont_form_recvd_date = ""
 
+		'Herehere
+		'TODO: Is this okay placement?
```
C-Love, line 23448: This is good placement for reading the information but the interaction should come after all the gathering of information. The interruption of the script operation shouldn't happening within the class because we likely will move this class to being available to any script. However, this means you will need a way to output the information about the review code and have it be a class parameter.

notes/eligibility-summary.vbs
```diff
@@ -23444,6 +23444,82 @@ class stat_detail
 		stat_mont_form_recvd_date = replace(stat_mont_form_recvd_date, " ", "/")
 		If stat_mont_form_recvd_date = "__/__/__" Then stat_mont_form_recvd_date = ""
 
+		'Herehere
+		'TODO: Is this okay placement? 
+		'TODO: Why does this only work if inside the sub?
```
C-Love, line 23449: A sub (subroutine) is a function within a class and is the only place that we should be 'doing' things other than saving information. So the sub for the stat_detail class reads all of the information from stat for the specific footer month that the instance of the class is created for. 

I know that is weirdly phrased. 
A class is a structure in code to create an object with certain traits. It has information saved in variables and actions as subs. It is something that we would use over and over again that needs to function the same with different details stored within each instance. An example is in our CLoS - that is based on a class that defines everything we want to know about each script file, name, category, policy, hot topics, tags, etc. We created the initial structure and then apply that to each script. 

For Elig Summ - the stat class is the detail of STAT for any given month and we have an array of stat_detail classes from each month. The structure and type of information they contain is the same but the content of each variable within the class will be different from month to month. 

This sub is kind of a constructor sub, but we call it explicitly instead of having it happen when initiated because we set some variables and that's how I knew to do it. BUT this sub is the way we fill all the variables with the information specific to the case for the month provided.

However, if we want to read more information during the sub, we need to create variables that are defined during this run for us to read outside of the sub. At line 18613 we start by defining these variables that are specific to the class. In order to use the variables outside of the class, we need to use 'dot notation'- ClassObject.VariableName  to identify the information. We use these in an array here so it actually looks like this: STAT_INFORMATION(month_count).footer_year

So, you can move all of this outside of the sub as long as you are using the dot-notation to call the variable. Since it doesn't seem that you are reading any new information this can probably all be moved out. REVW information is read on line 23378

notes/eligibility-summary.vbs
```diff
@@ -23444,6 +23444,82 @@ class stat_detail
...
+		'TODO: Why does it take so long to initially load the script?
```
C-Love, line 23450: It is a huge file with large objects to compile.

notes/eligibility-summary.vbs
```diff
@@ -23444,6 +23444,82 @@ class stat_detail
...
+		'TODO: Do we like how it double checks the value or do we not want to read the statuses twice (once above and once within the error handling). Could restructure:
```
C-Love, line 23451: No need to read statuses twice, they won't change. We should not include an exit here though. This is information gathering, we want to do things with that information later.

C-Love (commented):
The class structure is a whole new beast. There is a pretty good YouTube playlist by a channel called 'Simply Coded' and they have specific class video - https://www.youtube.com/watch?v=Hk60BXz-mps&list=PL72Es31dJnK6-ZFXXHFurNwJ6QtWPtxbz&index=53

PR #1766, opened 2024-03-29
1764 wage match op clean up

IlseFerris:
Removal of overpayment functionality from DEU - MATCH CLEARED. This has now been removed from all match-based scripts and we now have the just one overpayment functionality to work with. 

Instructions updated. Not great, but they will do. 

Will add policy and procedural references in separate pull. I have the CLoS out in another pull request.


courtrightd (approved):
Fixed one tiny typo, looks good! Approved!

Issue #1718, opened 2024-03-09
ADMIN - POLI TEMP MONTHLY UPDATES: enhance to read from Excel


IlseFerris:
Since we've aligned the ES POLI TEMP project to output, it will makes sense to align this automation to also use the same list. This will also be a user experience enhancement as well.  

Also change the poli look back period to less time. DHS has been updating poli temp with greater frequency.

IlseFerris:
Working as intended. Closing issue.

PR #1767, opened 2024-04-01
#1718 ADMIN - POLI TEMP MONTHLY UPDATES: Created bulk functionally

IlseFerris:
Was previously reference based supports, but now they are bulk and use the same list that the upload process also uses. Remove -6 month code, using -3 month eval, and that's in the funclib. Defaulted file path, and dialog supports for the file path and opening Excel. Added output updates in Excel vs. script end procedures.

I'm the sole user of this script, and I've tested this and it's working as intended.


MeganGeissler (approved):
Not sure if you want the msgbox on 234 for excel row or not. Otherwise looks good!

Issue #1752, opened 2024-03-26
Ex Parte Report update phase 1 list for SSI and person-based changes


courtrightd:
- [ ] Update ex-parte case selection to include all members with SSI income.
- [ ] Change to include all cases where at least 1 member meets ex-parte criteria, currently selects only cases where all members pass.

PR #1772, opened 2024-04-05
#1752 Ex Parte Phase 1 person-based and SSI changes

courtrightd:
Includes all cases in phase 1 that have SSI income.
Includes all cases in phase 1 that have at least 1 member passing ex-parte

### DIFF
--- a/admin/ex-parte-report.vbs
+++ b/admin/ex-parte-report.vbs
@@ -2608,7 +2608,7 @@ If ex_parte_function = "Prep 1" Then
 			case_has_no_income = ""
 			case_has_EPD = ""
 
-			appears_ex_parte = False			'we default this to true and find reasons that exclude the case from Ex Parte as we look at case data.
+			appears_ex_parte = False			'we default this to false, after reviewing all members we will set it true if any member passes ex parte
 			all_hc_is_ABD = True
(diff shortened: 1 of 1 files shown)


C-Love (changes requested):
The MEMBER_INFO_ARRAY is resized at line 2692, so we should add 'MEMBER_INFO_ARRAY(memb_appears_ex_parte, memb_count) = True 'Default this members ex_parte status to True' there as well. 

I think the section starting at line 2760 becomes dead code with the new logic, the case cannot appear ex parte if a person has not been found on the case. This ability to check for a person if no person is on the ELIG list should occur regardless. I think the 'If appears_ex_parte = True Then' condition should be removed and the whole section under 'If person_found = False Then' should be moved to right after the ELIG Connection is closed. The new parameter for the member array will need to be defaulted to true within this section as well. 

The new loop with the logic to find any member that appears ex parte would need to be after every possible person is reviewed for the case. 

Only other thought is if developer mode is on, add a script_end_procedure before the dealing with the excels - which is right now at 3244 - though maybe the connection closures should be moved up in that case as well. This is definitely optional since the dev mode is specific to you but there is automatic emails at this point.

C-Love (approved):
Everything looks good.

PR #1774, opened 2024-04-06
1773 ssn interface issue

IlseFerris:


dail/wage-match-scrubber.vbs
```diff
@@ -355,7 +356,7 @@ FUNCTION income_matrix(income_matrix_array, match_name, match_employer, quarter,
 		Text 370, 45, 30, 10, "Prosp"
 		Text 410, 45, 30, 10, "PIC"
 		Text 450, 45, 40, 10, "Quarterly HC"
-		'Displaying quarterly wage info for each jaeorb
```
C-Love, line 358: HAHAHAHA

C-Love (approved):
I couldn't really test but this all looks good. 

I want to be clear with what this looks like. It seems we are to the point that we no longer need to account for messages that have the SSN listed in them any more, we know that every message has the SSN removed and lists by member number only. And also that even if the SSN isn't displayed, each message can connect to the SSN to direct navigate to IEVS without having to enter the SSN. 

This change applies to all existing and new messages - to the DHS patch works on the old messages just as well as it does the new.

Basically they fixed what was broken when they changed the display and they fixed it in the background so they are all resolved. 

Also, there are lots of spelling errors that were fixed. Jaeorb is my favorite.

Issue #1764, opened 2024-03-29
DEU - MATCH CLEARED: Remove OP functionality


IlseFerris:
Separate functionality for overpayments is built into this script, and doesn't meet all the needs of the OP process as it doesn't have collectable vs. non-collectable supports and it doesn't CCOL note for multiple claims. 

These processes are not tied directly together, and having separate handling for setting up the claims makes the most sense for this process.

IlseFerris:
Closing issue. @MeganGeissler will add a issue to resolve the bug in the current functionality when an overpayment's claim number is needed.

PR #1775, opened 2024-04-06
#1764 CLoS: Updated policy & procedures for DEU - MATCH CLEARED

IlseFerris:
@C-Love that function in ALL SCRIPTS continues to be so helpful. Thank you!


C-Love (approved):
Looks great. Also way easier to read when you do it this way.

Issue #1770, opened 2024-04-04
Expedited Determination Report Script to be pulled back to AIT Operation


C-Love:
The process in which an OS team was running the Expedited Determination Report appears to have ended. 

Additionally, the spot checks process which is also supported in this script is not currently in operation. 

We need to update the script in two ways, 

- [ ] Restrict use to A&I Team only
- [ ] Remove the spot checks support process

C-Love:
This is complete and working. The script is now under complete control of the A&I Team.

PR #1778, opened 2024-04-11
#1770 Expedited Determination Report - Restrict to BZST and Update Interface

C-Love:
The Expedited Determination Report has previously used and supported by OS staff but this process ended in Jan. There was significant clean up of the data and Excel files needed. 

I have updated the script to ensure only BZST can run it and also allows the script to limit the run time. 

Currently the output to the Expedited Spot Checks is turned off until additional details can be reviewed.


IlseFerris (approved):
Makes sense to me. I approve thee!

Issue #1777, opened 2024-04-10
On Demand - BULK Applications - Report out Working List Size


C-Love:
In order to help understand the size of the work for On Demand, we are going to add the number of cases in the working list / data table to the end message so that we can review the number of cases when we look back at usage.

C-Love:
There might be an issue with On Demand where it ends if the MEMO glitches out.

C-Love:
We need a way for the function write_variable_in_SPEC_MEMO to tell that the script should NOT end if the MEMO is full but it should cancel the MEMO and should not attempt to add more lines to a MEMO. 

This is a challenge because this function is called in so many places. If we need to add more parameters, we will need to change each function call.

C-Love:
Updates completed. No errors reported.

PR #1780, opened 2024-04-11
#1777 Update On Demand End Message and Keep MEMO limit from failing BULK run

C-Love:
Need to review with whole team. Unsure if the FuncLib changes are the best option.


IlseFerris (approved):
I didn't test, but this makes sense to me. I approve thee!

MDR22 (approved):
Tested in MEMO and WCOM and it appeared to work as expected. Looks good!

Issue #1753, opened 2024-03-26
ADMIN-Review Report -- Update Review Report to unwind Interview Waiver


C-Love:
We need to start sending appointment notices for SNAP cases and not just MFIP cases. Update to Review Report to do this by the appt notice run in April (usually 4/25/24). 

NOMIs should not have to be changed as they are selected based on if an appointment notice was sent for the case but this should be confirmed. 

Additionally, no changes to individual scripts or supports for the application process should be needed because we did not change the notice process during the interview waiver for these cases.

MeganGeissler:
No changes were made, scripts was already updated. Branch closed without merging.

IlseFerris:
Line 3285 addresses the NOMI, but the APPT notices is where this code needs to be updated. Added @C-Love to the issue as well. 
@MeganGeissler can you look at Line 2514? And then we can remove all the PHE temp codes as well

MeganGeissler:
SNAP Waived interview extended through oct, 31st '24. Postponing this until then.

C-Love:
Notices are now sent for all SNAP and MFIP cases.

PR #1793, opened 2024-04-19
#1753 Start sending appt notices for SNAP  not just MFIP

MeganGeissler:
Commented out snap waived interview handling for appt notices
Activated standard appt notice handling (sends appt notices for SNAP not just MFIP


IlseFerris (changes requested):
I think this might have already been fixed. LIne 3286 is looking for MFIP to be active as an additional criteria.

MeganGeissler:
Pull not needed, this change was already implemented.

PR #1794, opened 2024-04-19
1663 Reported issues with accessing the mhc and deu menu

MDR22:
Reset dialog1 outside of declare_main_menu_dialog function call which may help to reduce dialog errors that pop up with Managed Care and DEU main menu scripts. Activated the links to script instructions for DEU scripts and updated SharePoint links to fix broken links.


IlseFerris (changes requested):
the instructions for REPT-IEVC List and REPT-INTR List are opening the general BULK script instructions. These will need updates, possibly in the CLoS. 

Everything else looks good and I approve this!

MDR22 (commented):
Updated Complete List of Scripts so that links to script instructions work for IEVC and INTR List scripts. Added changelog to DEU - MAIN MENU script

Issue #1776, opened 2024-04-08
NOTES - HC Evaluation Changes to Ex-Parte Y/N field in MAXIS


courtrightd:
The ExParte (Y/N) Field on STAT/REVW HC renewal pop-up is being removed. The field will be added to STAT/HCMI (person based) for each member.
HC Evaluation currently reads the status of this field, gives the worker update instructions and checks for correct updates. The field is used in 4 locations in the script.
Will need to evaluate how to update this logic to be person based.
Need to review current functionality and possibly create a temporary workaround to keep the script functional when the changes go live around April 15th.

PR #1795, opened 2024-04-19
#1776 Changes to ex-parte Y/N field on stat/revw and HCMI

courtrightd:
Updates NOTES - HC evaluation to read the new locations on STAT/REVW HC popup and check STAT/HCMI for potential errors.

Updates wording in dialogs to guide HSRs with coding of review/HCMI.

Also fixes ADMIN - Training case creator to write into new field locations.

Changes drop-down options and noting for reasons a case will fail phase 1. This is a partial solution while full support of asset policy is created.


C-Love (changes requested):
I have some concerns about the HCMI review month being separate and potentially mismatching the month listed on  REVW, but I think we will just have to see how MAXIS actually behaves and make adjustments. 

I think the navigation in the new function won't work quite right, on line 240. The other comment is just on verbiage.

C-Love (approved):
Looks like that will work. I think this is good to go.

Issue #1769, opened 2024-04-03
SPEC/MEMO Remove obsolete info and revise digital space verbiage


MeganGeissler:
Revisions include: 
-	Remove obsolete location: Hopkins
-	Replace “all interview are completed via phone” to “can be completed via phone”
-	Rework the digital space verbiage to move away from PHE messaging 
-	Increase readability 

Actions to take: 
- [x] Draft approved by AT & MD
- [x] Update required files 
- [x] Update word docs in this folder and on SharePoint: Eligibility Support\Restricted\QI - Quality Improvement\BZ scripts project\Projects\Completed projects\On Demand Waiver\Notices Verbiage
- [x] Ask JF if required to have all service centers on notices

Files require updates:
- [x] Func Lib: 4 Appt. Notices & NOMIs
- [x] Interview no show (1) ("retired script")
- [x] Eligibility notifier (1) - commented out already
- [x] Bulk recertifications (20) - commented out already
- [x] Individual appointment letter (1) (retired script)
- [x] Individual nomi (1) (retired script)
- [x] Individual recertification notices (10) (retired)
- [x] On demand waiver recertifications (20) (retired)
- [x] Testing notices (24) -- Casey's test file
- [x] New new on demand waiver applications (2) -- Casey's test file
- [x] New on demand waiver applications (2)-- Casey's test file
- [x] On demand waiver applications send manual nomis (1) -- Casey's test file
- [x] On demand waiver app part two (2) -- Casey's test file

Memo Docs on Sharepoint- do not change name

- [x] https://hennepin.sharepoint.com/:w:/r/teams/hs-economic-supports-hub/_layouts/15/Doc.aspx?sourcedoc=%7BF763F819-447A-4EAE-A8C8-B6778CE27402%7D&file=Application%20Appointment%20Notice.docx&action=default&mobileredirect=true&DefaultItemOpen=1
- [x] https://hennepin.sharepoint.com/:w:/r/teams/hs-economic-supports-hub/_layouts/15/Doc.aspx?sourcedoc=%7B12731A19-59EE-422E-A43E-BA0CD9A2D28F%7D&file=Recertification%20NOMI.docx&action=default&mobileredirect=true&DefaultItemOpen=1
[...]

PR #1802, opened 2024-04-25
#1769 Revised Appt notice and nomi functions in funclib

MeganGeissler:
Revised appointment notice and nomis in func lib functions. I didn't run into any issues with the memos being too long but always good if someone else can verify this too!
- Removed obsolete hopkins locations
- Added additional locations 
- Eliminate PHE verbiage, updated to be more current


courtrightd (approved):
Tested good on this end!

MDR22 (approved):
Looks good for me too!

IlseFerris (approved):
Looks great, I approve thee!

Issue #1663, opened 2024-01-30
Reported issues with accessing the MHC menu


C-Love:
Unclear if this is an issue in the code of the menu or somewhere else in the line of functionality. 

Discovery and testing to be completed.

C-Love:
No additional reports, this might have actually been due to VPN issues that were happening at the same time.

MDR22:
Received new report of dialog issues with DEU - MAIN Menu script so reopening issue. Also, will activate links to DEU script instructions.

PR #1808, opened 2024-04-29
#1663 Change script file names and script name within script to align with CLOS

MDR22:
Updated CLOS with updated script names but did not change actual script file names. Did this to ensure that DEU main menu links to correct instructions for IEVC and INTR list scripts.


IlseFerris (approved):
Worked for me, I approve thee!

PR #1814, opened 2024-05-08
#123 NOTES - Elig Summ - XFS clarity - better non-coop info

C-Love:
Updates to SNAP functionality to be cleaner:

- Better understanding of Expedited packages - to not include months that will not issue.
- More detail around Non-coop case test failures. This also applies to DWP and MFIP
- Update to document that shelter expense has no limit if the allowed shelter expense is over the MAX shelter amount.

Additional updates will be coming but these should be stable now.


C-Love:
I have some more updates to pull in first.

PR #1817, opened 2024-05-10
1813 func lib addr hest shel function

MeganGeissler:
Updated Coordinates and updated format/handling issues. 
-Updated phone number coordinates for new layout 
-Revised State list to remove dash -- this allowed the variable to populate the dialog if a value existed 
-Mail_state and resi_state leverages state list 
-Updated addr_eff_date handling so it has xx/xx/xx format 
-Updated mail_state handling so original and current matched -- otherwise it was updating addr when there wasn't an update
-Revised handling for writing address on 1 or 2 lines

### DIFF
--- a/MASTER FUNCTIONS LIBRARY.vbs
+++ b/MASTER FUNCTIONS LIBRARY.vbs
@@ -1541,9 +1541,8 @@ function access_ADDR_panel(access_type, notes_on_address, resi_line_one, resi_li
 		mail_street_full = trim(mail_line_one & " " & mail_line_two)
         mail_city = replace(mail_city_line, "_", "")
 		Call get_state_name_from_state_code(mail_state_line, mail_state, TRUE)		'This function makes the state code to be the state name written out - including the code
-		mail_state = replace(mail_state_line, "_", "")
+		mail_state = replace(mail_state, "_", "")
 		mail_state_line = replace(mail_state_line, "_", "")
-		mail_state = trim(mail_state)
 		mail_state_line = trim(mail_state_line)
 		mail_zip = replace(mail_zip_line, "_", "")
 		If text_yn_one = "_" Then text_yn_one = ""								'Changing blanks to nulls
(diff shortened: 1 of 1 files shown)


MASTER FUNCTIONS LIBRARY.vbs
```diff
@@ -1540,9 +1540,12 @@ function access_ADDR_panel(access_type, notes_on_address, resi_line_one, resi_li
         mail_line_two = replace(mail_line_two, "_", "")
 		mail_street_full = trim(mail_line_one & " " & mail_line_two)
         mail_city = replace(mail_city_line, "_", "")
-        mail_state = replace(mail_state_line, "_", "")
-        mail_zip = replace(mail_zip_line, "_", "")
-
+		Call get_state_name_from_state_code(mail_state_line, mail_state, TRUE)		'This function makes the state code to be the state name written out - including the code
+		mail_state = replace(mail_state_line, "_", "")
```
C-Love, line 1544: This will replace the assignment to this variable that we just had on the previous line.

MASTER FUNCTIONS LIBRARY.vbs
```diff
@@ -1540,9 +1540,12 @@ function access_ADDR_panel(access_type, notes_on_address, resi_line_one, resi_li
...
+		mail_state_line = replace(mail_state_line, "_", "")
+		mail_state = trim(mail_state)
```
C-Love, line 1546: This should be unnecessary with the function call.

C-Love (changes requested):
I only have some concerns about the state code handling. Everything else looks good.

MASTER FUNCTIONS LIBRARY.vbs
MeganGeissler: Whoops- good catch. I meant that to be mail_state for both var in that line but based on your comment below I'll plan to remove the line anyhow!

C-Love (approved):
Looks good!

PR #1819, opened 2024-05-14
#123 NOTES - Elig Summ - SNAP Enhancements

C-Love:
Updates to Eligibility summary:
1. Better handling of Expedited in CASE/NOTE. Making sure the header only lists the months that are approved and the benefit amount is only for approved months. This is primarily for cases that will close SNAP due to postponed verifications, MAXIS looks kind of weird in these situations.
2. Adding a display of NON-Cooperation into CASE/NOTE from detail in STAT/DISQ. This also applies to MFIP and DWP cases. 
3. Update to the display of Shelter Expense for cases that are approved and exceed the MAX shelter expense to indicate that there is no limit to this expense deduction.
4. Added different information to the SNAP closure CASE/NOTE for cases in a SNAP to MFIP transition.
5. Created the ability to read cases that are indicated as homeless and have a verified shelter expense and the expense amount is missing or under the limit listed in CM 0018.15 and create a txt file to review cases that meet this criteria.

### DIFF
--- a/notes/eligibility-summary.vbs
+++ b/notes/eligibility-summary.vbs
@@ -69,6 +69,23 @@ call changelog_update("07/05/2022", "Initial version.", "Casey Love, Hennepin Co
 changelog_display
 'END CHANGELOG BLOCK =======================================================================================================
 
+' REVIEW BLOCK --- These are the possibilities for completing QC Reviews -------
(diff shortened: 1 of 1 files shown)


IlseFerris (approved):
Looks good to me. I approve thee!

PR #1820, opened 2024-05-17
1789 notes returned mail, update to reflect unclear info

MDR22:
Updated returned mail to reflect findings from FNS audit to properly act on mail returned as undeliverable (mail returned without forwarding address). Per FNS audit: when the county receives returned mail without a forwarding address for simplified reporting households, the county should hold that information and follow-up at the next scheduled certification action or periodic report about the potential change in residence. If mail returned without a forwarding address, then script will determine if case is an active, SNAP-only, 6-month reporting and if so, it will CASE/NOTE that worker should follow up on address change at renewal/recertification and not take negative action.

Will update script instructions once finalized.


C-Love (approved):
Looks good. Worked as expected.

Issue #1823, opened 2024-05-22
Update NOTES - CAF FSET and ABAWD codes


IlseFerris:
Code 23 - Pregnant is missing in the FSET/WREG Status selections is missing.
Code 13 - ABAWD Banked Months in the ABAWD Status selections is missing.
All other codes are correct and present.

PR #1826, opened 2024-05-23
#1823 NOTES - CAF: updated FSET and ABAWD statuses

IlseFerris:
Added PX and ABAWD Banked Months codes to applicable drop lists. Added changelog.


courtrightd (approved):
Looks good, tests good. 
The party approves, comrade.

Issue #1816, opened 2024-05-09
Client Contact NAV Error - if worker navigates into a CASE/NOTE


C-Love:
If a worker navigates into a CASE/NOTE when the dialog is displayed the script does not correctly create a new CASE/NOTE because it can't start one. All kinds of errors. 

Update the dialog loop to make sure we are NOT in a CASE/NOTE when the dialog is exited.

PR #1827, opened 2024-05-23
#1816 Client contact NAV error if in a case note

courtrightd:
Fixes issue 1816, gets the case out of wherever it may be before starting a new case note.


IlseFerris (approved):
I approve thee!

Issue #1824, opened 2024-05-22
DEU - MATCH CLEARED: Add handling for matches with missing worker information


IlseFerris:
Even if a case is in our county if the worker information on the INFC/IEVP menu is missing, the match cannot be cleared. Current messaging says that the case is out-of-county which is not particularly helpful.

PR #1829, opened 2024-05-23
#1824 DEU - MATCH CLEARED: Updated messaging for blank co workers

IlseFerris:
Added changelog
Changed to cancel w/o confirmation
Added if statement if out of county code is found to check for blank county name and if found gives the user a different script end procedure message.


courtrightd (approved):
Request approved, your grace. 
Long live the queen!

PR #1832, opened 2024-05-23
1822 error reported with work# url

MDR22:
Updated Work# link for speciality and Hennepin power pads


IlseFerris (approved):
Well I can't really tell but I approve thee!

Issue #1194, opened 2023-03-02
Retire NOTES - Approved Programs


C-Love:
Eligibility Summary is replacing NOTES - Approved Programs.

C-Love:
Approved Programs functionality is no longer available. Next steps for this work #1858

PR #1833, opened 2024-05-23
#1194 NOTES - Approved Programs - Final sundown message

C-Love:
Changelog for final sundown - WITH ASCII ART.

Also updated some of the information and functionality in the Elig Summ redirect.


courtrightd (approved):
The art is so real I can feel the suns rays from here. 11/10, would approve.

PR #1838, opened 2024-05-23
#1744 Fixes date variable to ensure it writes to PROG

courtrightd:
Corrects an issue with a blank interview date in the dialog causing string error. If the user removes it from the dialog and PROG does not have an interview date, will now enter today's date automatically.


C-Love (approved):
Works like a dream. Glad you found it!

Issue #1807, opened 2024-04-28
ACTIONS - EARNED INCOME BUDGETING: Add editing check


IlseFerris:
HSR, AG-K, reported the following bug to ACTIONS - EARNED INCOME BUDGETING: 

> On the previous run of this script I forgot to REIN the case first so it could actually be updated but the script ran like it was updating, and case noted. I cleared the second case note so there's no longer an extra one but it did case note the first time when it couldn't actually update anything.

Based on this feedback we will update the script to ensure that editing can indeed take place.

C-Love:
This appears to be working.

PR #1839, opened 2024-05-23
#1807 ACTIONS - EIB - check to ensure panels can be updated

C-Love:


courtrightd (approved):
Tested great! Approved!

Issue #1813, opened 2024-05-02
FuncLib - access_ADDR_panel "WRITE" function - correct phone number coordinates 


MeganGeissler:
Phone number coordinates are incorrect for the "WRITE" portion of this function when it clears and tries to write a new value. 
16, 33 is where phone number 1 starts, funclib currently has it starting at 16, 39

PR #1840, opened 2024-05-23
#1813 Updated ADDR WRITE FUNCT- to handle for either 1 or 2 street address fields in a dialog

MeganGeissler:
Issue with recent update to Access_Addder "WRITE" portion. I had adjusted the script to use resi_street_full and mail_steet_full when writing however this didn't handle for the fact that dialogs sometimes have 1 or 2 fields to capture the street address. 

Updated handling to handle for both scenarios. Extra testing on this would be appreciated :). Return Mail and LTC 5181. Use had reported issue with 5181 not writing correct street address which is brought this to our attention.

### DIFF
--- a/MASTER FUNCTIONS LIBRARY.vbs
+++ b/MASTER FUNCTIONS LIBRARY.vbs
@@ -1716,6 +1716,7 @@ function access_ADDR_panel(access_type, notes_on_address, resi_line_one, resi_li
 	        Call create_mainframe_friendly_date(addr_eff_date, 4, 43, "YY")
 			
 			'This functionality will write the address word by word so that if it needs to word wrap to the second line, it can move the words to the next line
+			move_to_line_two = FALSE
 			If resi_line_one = "" AND resi_line_two = "" Then 
 				If len(resi_street_full) > 22 Then 
 					resi_words = split(resi_street_full, " ")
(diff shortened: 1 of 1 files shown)


C-Love (approved):
Hold on - There is an issue with the word count in the lines

MASTER FUNCTIONS LIBRARY.vbs
```diff
@@ -1715,26 +1716,52 @@ function access_ADDR_panel(access_type, notes_on_address, resi_line_one, resi_li
...
-	        End If
+			If resi_line_one = "" AND resi_line_two = "" Then 
+				If len(resi_street_full) > 22 Then 
+					resi_words = split(resi_street_full, " ")
+					write_resi_line_one = ""
+					write_resi_line_two = ""
+					For each word in resi_words
+						If write_resi_line_one = "" Then
+							write_resi_line_one = word
+						ElseIf len(write_resi_line_one & " " & word) =< 22 Then
```
C-Love, line 1727: If a smaller word is after a longer word - the order or words will get messed up. 

'1135 Fifth Avenue Apartment 1905' became 

'1135 Fifth Avenue 1905'
'Apartment'

We need a boolean to identify if we are moving to line 2

MASTER FUNCTIONS LIBRARY.vbs
```diff
@@ -1715,26 +1716,52 @@ function access_ADDR_panel(access_type, notes_on_address, resi_line_one, resi_li
...
+			Else
+				resi_street_full = trim(trim(resi_line_one) & " " & trim(resi_line_two))
+				If len(resi_line_one) or len(resi_line_two) > 22 Then 
+					resi_words = split(resi_street_full, " ")
+					write_resi_line_one = ""
+					write_resi_line_two = ""
+					For each word in resi_words
+						If write_resi_line_one = "" Then
+							write_resi_line_one = word
+						ElseIf len(write_resi_line_one & " " & word) =< 22 Then
```
C-Love, line 1750: This would create the same issue

C-Love (changes requested):
The update worked but we have the possibility of messing up the word order.

courtrightd (approved):
Testing good!

MASTER FUNCTIONS LIBRARY.vbs
MeganGeissler: @C-Love - I've added boolean for resi and mail handling. Take a peek and let me know what you think!

C-Love (approved):
Worked great. Logic is solid. I approve.

PR #1842, opened 2024-05-28
#1776 Updates for ex-parte person based checks of revw/hcmi, mixed household handling

courtrightd:


C-Love (approved):
Looks good. I ran it to the point that I would have had to update the case and it worked well. Didn't go through to database update or CASE/NOTE but those look solid. Plus this team will fairly reliably report issues. Good to go.

Issue #1844, opened 2024-06-03
ACTIONS Transfer Case - Script requires excluded time if GRH = Active 


MeganGeissler:
User JD reported that for GRH, community setting isn't excluded time. Verified w/ CM06.18: https://www.dhs.state.mn.us/main/idcplg?IdcService=GET_DYNAMIC_CONVERSION&dDocName=cm_000618&RevisionSelectionMethod=LatestReleased

Script currently has error handling that requires excluded time = yes if GRH = Active or App Open.

PR #1845, opened 2024-06-03
#1844 Removed excluded time req for GRH in error handling

MeganGeissler:
Removed excluded time = yes req for active GRH in error handling in response to JD's email. 

Script will still default exclude time to Yes if GRH = active/app open however it will allow the user to change it to NO for circumstances that are not excluded time such as community settings.


MDR22 (approved):
We chatted through this, and I think the changes makes sense. I'll defer to others though.

C-Love (approved):
Looks good, no concerns here.

PR #1852, opened 2024-06-04
#1194 Approved Programs always Redirects to Elig Summ

C-Love:
Updated Approved Programs to always redirect to Eligibility Summary

Func Lib changes are because the ASCII art in Approved Progs has  | - which is used as a delimiter and it is making all of the changlog messages show up.


IlseFerris (approved):
Worked as intended. I approve thee!

Issue #1853, opened 2024-06-05
NOTES-CLIENT CONTACT, allow attempted interview for voicemail option


MDR22:
Received email from worker that client contact prevents use of word "interview" even when interview attempted for voicemail option. Update to allow use of interview for voicemail option.

PR #1854, opened 2024-06-05
#1853 Allow 'interview' for voicemail option if interview attempted

MDR22:
Per email in team email. Changed error handling to allow worker to select "Voicemail" contact type option and indicate a phone interview was attempted. Previously only the "Phone call" option would allow the checkbox to be checked, but it makes sense that leaving a voicemail might be another way to indicate the same information as a phone call for an attempted phone interview.


IlseFerris (approved):
Makes sense to me, I approve thee! 
Please add a changelog.

C-Love (approved):
It works! Looks good. I agree with Ilse, just a changelog and its good to go.

MDR22:
I updated to include the changelog

PR #1857, opened 2024-06-05
#1865 Remove ES-CasesPending Table Input

C-Love:
Same block of code removed from all. Had to add validate_MAXIS_case_number to subsequent app because that is where the MAXIS_case_number variable is trimmed. This was a problem at one point and we want to make sure it gets trimmed. 

Did not update any scripts in Test folders.


IlseFerris (approved):
Looks good. I approve thee!

PR #1865, opened 2024-06-11
1729 shelter permanent housing revise

MeganGeissler:
Limited revisions to the scope of the request - remove self pay. A bigger overhaul of shelter scripts will happen in the future. Changes made include:
-Eliminated self-pay related fields: Revised main dialog error handling and case note to align 
-Added initial dialog to capture case# so that we can pull in phone numbers from the ADDR panel. 
-Change log and closing documentation added script end procedure


shelter/permanent-housing-found.vbs
```diff
@@ -51,150 +52,198 @@ changelog_display
...
-		If isNumeric(TANF_months) = False then err_msg = err_msg & vbNewLine & "* Enter numeric number of TANF months used."
-		If isNumeric(num_days_shelter) = False then err_msg = err_msg & vbNewLine & "* Enter the numeric number of days in shelter."
-		IF worker_signature = "" THEN err_msg = err_msg & vbCr & "* Sign your case note."
-		IF err_msg <> "" THEN MsgBox "*** NOTICE!!! ***" & vbNewLine & err_msg & vbNewLine
-	LOOP UNTIL err_msg = ""
+	  DO
+      err_msg = ""
+      Dialog Dialog1
+      cancel_confirmation
+      IF IsDate(move_date) = False then err_msg = err_msg & vbNewLine & "* Enter a numeric move in date."
```
MDR22, line 149: For error handling for the dates, I think it might be clearer to say something along the lines of "Enter a move in date in the format MM/DD/YYYY" instead of numeric date.

shelter/permanent-housing-found.vbs
```diff
@@ -51,150 +52,198 @@ changelog_display
...
+      If isNumeric(monthly_rent) = False then err_msg = err_msg & vbNewLine & "* Enter the monthly rent."
+      If isNumeric(num_days) = False then err_msg = err_msg & vbNewLine & "* Enter a numeric number of days in shelter."
+      If isNumeric(rent_needed) = False then err_msg = err_msg & vbNewLine & "* Enter a numeric rent amount needed to move in."
+      If isNumeric(DD_needed) = False then err_msg = err_msg & vbNewLine & "* Enter a numeric DD amount needed to move in."
+      If rent_subsidy_paid = "" then err_msg = err_msg & vbNewLine & "* Enter who will be paying for the rental subsidy."
+      If isNumeric(rent_subsidy) = False then err_msg = err_msg & vbNewLine & "* Enter a numeric subsidy amount."
+      If additional_funds = "" then err_msg = err_msg & vbNewLine & "* Enter who will be issuing the additional funds."
+      If LL_name = "" then err_msg = err_msg & vbNewLine & "* Enter the Landlord's name."
```
MDR22, line 157: It might be helpful to include trims for the required fields so that someone can't enter a space in a field to skip it.

MDR22 (approved):
Tested the script and it worked as expected. Had a few minor comments about error handling. Looks good!

shelter/permanent-housing-found.vbs
MeganGeissler: good point on both accounts - I was moving too fast. Thanks!

IlseFerris (changes requested):
Not part of this issue but for the client_phone functionality to work the dialog will need to be in the dialog do...loop. Otherwise if any mandatory fields are missed and a user types a phone number vs. selects one, it will blank out. See NOTES - CLIENT CONTACT for an example. 
I pushed a couple of small updates too.

IlseFerris (approved):
Works great, I approve thee!

PR #1867, opened 2024-06-12
1145 es workflow task based dail capture updates

IlseFerris:
Logic for which messages are selected is commented and also in the issue itself. 
I recommend changing date to test_date to test. 
@C-Love just confirmed that everyone has access to these data tables. Deleting the table has no effect if tested prior to 6 PM.


MeganGeissler (approved):
Looks good!

Issue #1869, opened 2024-06-13
Function - find_last_approved_ELIG_version - the 99 code doesn't always work?!


C-Love:
We need an option in the find_last_approve_ELIG_version that does not use the code 99 on the command line. 

This has only been found in MSA so the functionality is specific to MSA at this point. 

Need to push this through quickly as it is an issue with Elig Summ.

C-Love:
New functionality in place to support MSA Elig. No news on this being broken or having any unintended impact.

PR #1870, opened 2024-06-13
#1869 FuncLib - find_last_approved_ELIG_version - MSA 99 code might not work

C-Love:
I tested this on case 2229996 when used in Elig Summ.


IlseFerris (approved):
Looks good. I approve thee!

PR #1871, opened 2024-06-15
#123 NOTES - Elig Summ - a number of enhancements and handling for specific situations

C-Love:
These updates for strengthen the functionality and handling of Eligibility Summary.

- Only make sure that we are entering the right case notes for cash cases, there have been instances of case notes that are in disagreement or duplicate notes.
- Add family wage level amount to the cash deny note.
- Better phrasing the cash withdraws when processed with PACT. 
- Read verifs from STAT to autofill information for Cash Deny
- Keep from duplicating WCOM lines. 

Next update will include additional edit boxes to explain case actions (FIAT, Proration, MFIP Vendor)


IlseFerris (approved):
Make sense to me. I approve thee!

PR #1872, opened 2024-06-18
#123 elig summ MFIP Vendor and phrasing updates

C-Love:
Added the ability to document vendor payments as a part of the MFIP approval and eligibility.


courtrightd (approved):
Should probably add a changelog. There are typos in the previous changelog entry while you're in there...
Tested fine, note looks good.

Issue #1792, opened 2024-04-19
All cases will be at standard policy in July - Remove Clarifying CASE/NOTE


C-Love:
We had a process in place to enter a CASE/NOTE on cases that should have standard HC policy apply to them. This was supportive of the unwind process since standard policy applied to cases at different time over the 15 months of the unwind. 

This functionality is in:

- [x] NOTES - HC Evaluation
- [x] ADMIN - Review Report
- [x] NOTES - Eligibility Summary

We should be completely done using this functionality by the end of June. All cases in July and onward should have standard HC policy applied, PHE based continuous coverage will no longer apply to any case. There should be no reason for this policy to be applied again in the future and we can just remove it all.

C-Love:
This is done - no errors reported.

PR #1879, opened 2024-06-27
#1792 - ADMIN - Review Report - remove HC Standard Policy CASE NOTE

C-Love:
Removing all coding around creating return to standard HC case notes in ADMIN Review Report. Will update HC Eval once it is not being updated. 

I will be testing this through on the 1st as a part of routine runs.


IlseFerris (approved):
Not tested but looks good. I approve thee!

Issue #1473, opened 2023-09-05
Replacing Stolen EBT Food


IlseFerris:
New procedures in POLI TEMP have been added effective 09/2023 for how an eligibility worker would support a request for replacing stolen EBT food benefits. 

The procedure is detailed and fussy and would benefit from automation and script guidance through the process. What happens if eligibility staff don't fully complete each step of the process, a back and forth via SIR email occurs, which could result in residents getting degrade or no customer service. 

This functionality could be merged with NOTES - DISASTER FOOD REPLACEMENT to create a comprehensive SNAP replacement support. 

Refer to the following POLI TEMP references: 
- TE.02.11.18 Food Destroyed in a Misfortune or Disaster
- TE02.11.127 Replacing Stolen EBT Food - Client Requests
- TE02.11.126 Replacing Stolen EBT Food - Client Notified
-

MDR22:
Megan and I reviewed each of the POLI TEMP references and the current NOTES - DISASTER FOOD REPLACEMENT script.  Identified opportunities to combine processes into a single script due to overlap in process steps. I plan to work on further planning out and developing script later in November. Background materials and files saved here: T:\Eligibility Support\Restricted\QI - Quality Improvement\BZ scripts project\Projects\Unclear Information

MDR22:
Reviewed proposed approach to issue. Plan to create a script that incorporates the Food Destroyed in Disaster script into a script with the Food Benefits Stolen process. Will include options to document steps in process to allow for multiple workers to work through request - same as current Food Destroyed script. Inquired with DHS regarding lack of SPEC/MEMO direction for Food Destroyed in Disaster in POLI/TEMP, will update script based on direction received.

MDR22:
RE

> Reviewed proposed approach to issue. Plan to create a script that incorporates the Food Destroyed in Disaster script into a script with the Food Benefits Stolen process. Will include options to document steps in process to allow for multiple workers to work through request - same as current Food Destroyed script. Inquired with DHS regarding lack of SPEC/MEMO direction for Food Destroyed in Disaster in POLI/TEMP, will update script based on direction received.

Received response from DHS. SNAP policy currently working on updates to Food Destroyed in Misfortune/Disaster (TE02.11.18) that will include information about sending notices. DHS noted that MAXIS will send notices when denial reasons are provided (TE02.11.07) but this may be outside the scope of current script.

IlseFerris:
[...]

PR #1883, opened 2024-07-01
#1473 NOTES - EBT STOLEN BENEFITS, correct script name

MDR22:
Corrected script name in complete list of scripts, which had been causing errors when users tried to access script


courtrightd (approved):
Approved!

PR #1891, opened 2024-07-02
#1792 NOTES - HC Eval - Remove handling for NOTE about return to standard Policy

C-Love:
To support the HC PHE unwind, we were identifying cases that met standard policy by adding a CASE/NOTE. 

Now all cases should have standard policy applied and we no longer need to identify when that occured.


courtrightd (approved):
Looks good from here, approved!

PR #1897, opened 2024-07-08
1776 HC Eval ex parte multi-member-households

courtrightd:
Fixes ex parte phase 1 evaluation to function when there are more than 2 members on healthcare.


C-Love (approved):
I found some formatting updates that were easy enough to do so we could get this moved on. Tested it on a case I found.

Issue #1899, opened 2024-07-09
ELIGIBILITY SUMMARY - MAXIS Access to MONY is not Universal


C-Love:
Error reported by worker because they do not have access to MONY/INQX. 

Discussed with the worker and supervisor and because TE and SH are both METS workers that have moved to MAXIS HC, they did not have standard MAXIS training of other MAXIS workers. We can request this access but may have issues. Their supervisor NH is requesting this access.

In the meantime, we are going to create a workaround in Eligibility Summary to ignore programs other than HC for these workers. This will provide the ability for them to use the script in all situations while still reviewing all of the programs for other workers. MONY view is important for MFIP, DWP, GRH, and EMER.

C-Love:
Done - confirmed by TE

PR #1900, opened 2024-07-09
#1899 NOTES - Elig Summ - MONY access workaround

C-Love:


courtrightd (approved):
This looks like it should work!

Issue #1868, opened 2024-06-13
Update On Demand Applications to ignore Cash/SNAP interview mismatch


C-Love:
The current On Demand process flags any cases where there is a mismatch between SNAP and Cash as to if an interview has been completed or not. This process is to support a single interview issue. With the current SNAP Interview waiver if a SNAP case has an interview date and there is no Cash interview date, this could be correct processing. 

We need to be sure to check some cases to ensure we do not add to disparate treatment.

C-Love:
Update - upon review of these cases, we need to review these cases. However, they are put on the list daily and have to be reviewed daily. We need to see if there is a way to identify when this issue was reviewed and then reassign the case for review after a set amount of time (a week maybe). 

This functionality will need:

- Specific notes entered into the table and carried through to identify when a review is done. 
- Special functionality when the dashboard is run to complete this type of review.
- The ability to blank this 'Next Action Needed' so that all of the parts of the functionality operates. 
- We are limited because we cannot update the structure of the data table. 

This will be challenging logic to figure out. 

Moving to Q3 to be reviewed in July.

C-Love:
This is working, no issues reported.

PR #1902, opened 2024-07-11
#1868 ADMIN - On Demand Dashboard - Update worklog directly from SQL - no Array

C-Love:
This functionality is just to ensure the completion of the End of Day process doesn't break entirely if the connection is lost. Using the array to bridge the gap offers a chance for the end of day functionality to fail in the middle. This might cause the need to restart the functionality but no information would be lost. 

We are basically just mitigating interrupted runs due to worker error or connection issues.


courtrightd (approved):
Looks good, approved comrade!

Issue #1905, opened 2024-07-12
ACTIONS - Transfer Case - error if spaces are entered with x-number


C-Love:
User (AL) reported an error when entering the x-number into the initial dialog. She reported the error message 'Update transferring Service Worker'. this err_msg is associated with the length of the variable and it looks like this is missing a trim which could cause the script to count spaces. 

Add a trim to the first dialog handling for the x-number.

C-Love:
This is resolved - no additional errors have been reported.

PR #1906, opened 2024-07-12
#1905 ACTIONS - Transfer Case - trim x-number

C-Love:
Error reported if a space is in the field with the x-number as it counts the characters incorrectly.


MeganGeissler (approved):
Looks good to me!

courtrightd (approved):
Looks good, approved!

Issue #1910, opened 2024-07-16
NOTES - EMERGENCY, unable to navigate to CASE/NOTE


MDR22:
Worker reported that script was unable to navigate to CASE/NOTE from the “Date EA Last Used by Person” panel. Update script to PF3 out of panel before navigating to CASE/NOTE.

PR #1911, opened 2024-07-16
#1910 NOTES - EMERGENCY & Function navigate_to_MAXIS_screen, add handling for navigating away from EA history panel

MDR22:
Update handling for function navigate_to_MAXIS_screen to include handling for EA history panel screen.


IlseFerris (approved):
Makes sense, I approve thee!

Issue #1885, opened 2024-07-01
ELIGIBILITY SUMMARY - Enhance for Health Care Approvals


C-Love:
Additional support for Health Care in Eligibility Summary needs to be added.

Health Care process and procedures are being evaluated and reviewed. As these processes are clarified, we need to align Eligibility Summary functionality to these supports. 

There may be some tweaks to the detail in dialog and CASE/NOTE to make this script handle for all the intricacies of Health Care:

- [ ] Better handling for Closed or Denied in the CASE NOTE header - currently this is disabled because it is not functioning well. 
- [ ] Review TEFRA Note and Information Needed.
- [ ] Updates to 1619 Approvals and Details.
- [ ] Sage/BC approval information updates.
- [ ] Better handling for Method X eligibility.
- [ ] Some ineligible approval packets are not grouping correctly.
- [ ] Better PACT Handling.
- [ ] Change the HC Header when eligibility is moved to METS.
- [ ] Review RETRO handling and information display.
- [ ] Review detail around HC cases where resident is in a facility.

Additional information may come up as we get the Health Care section and workers established.

PR #1916, opened 2024-07-24
#1885 NOTES - Elig Summ - Bug when HC Retro request is before app date

C-Love:
This update allows the script to continue if STAT does not exists for an approval month because Health Care allows for approvals for months prior to the app date. No handling for other programs.


IlseFerris (approved):
Nothing looks out of place. Didn't run, but sounds like you have a tester ready to go. Maybe throw a changelog in though. I approve thee!

Issue #1924, opened 2024-08-06
ELIGIBILITY SUMMARY - Bug on MSP HC programs - verif edit boxes missing


C-Love:
The script is not showing the verification date or explanation edit boxes are not being displayed for the MSP program. 

Need update to the dialog to display the edit boxes correctly.

C-Love:
Confirmed this is working with CK.

Issue #1923, opened 2024-08-06
ELIGIBLITY SUMMARY - Formatting of MF Special Diets needs adjustments


C-Love:
There are some cases in which the diets do not overlap. The clarification of the amounts needs to be reviewed and updated. Reported by RK.

C-Love:
This appears to be done and working well.

PR #1925, opened 2024-08-06
#1923 & #1924 Elig Summ - overlap/missing diets verbiage & msp verifs fail

C-Love:
The Elig Summ updates here are for:
1. Formatting of MFIP Special Diet NOTE - there are some oddities with diet overlap vs diet information that does not fit on the panel. 
2. There is a missing field for Medicare Savings Programs for verifications failed date and explanation.

There will be an additional pull for supporting MFIP Special Diets for multiple members on a single case but I want to get these out here quickly


courtrightd (approved):
Fixed 1 spelling issue. Looks a-ok, approved.

C-Love:
Great! Thank you!

Issue #1811, opened 2024-04-30
NOTES - Application Received Update Case Transfer process for GRH cases


C-Love:
The current process for transfers in Application Received that supports the Case Assignment process is limited in how it handles GRH cases. We need additional clarification of process updates to enhance the script, but any process can only be implemented if the logic is in the script. 

Current proposal to create a list of GRH Maintenance Caseloads that we do not transfer GRH cases out of when pending program requests. There is also a need to create a separation between adult and family GRH cases. Review and approval of this process is needed before it can be implemented.

C-Love:
This work is completed and confirmed.

PR #1929, opened 2024-08-08
#1811 - App Recvd - GRH transfer Process Update

C-Love:
Clarified the logic for App Received when dealing with GRH cases. These cases should not be transferred as often and we need a split between adults and families.


MDR22 (approved):
Looks good to me!

C-Love:
Thank you both!

PR #1930, opened 2024-08-12
#1923 NOTES - Elig Summ - MF Spec Diet allow multiple Members

C-Love:
We cannot easily separate the checks because they are not associated with a specific member.

The resolution is to list all of the MF Special Diet details for all members and combine the checks.


courtrightd (approved):
Looks good to me!

PR #1934, opened 2024-08-13
1464 utilities all scripts and clos hyperlink functionality for te resources

MeganGeissler:
Updated Utilities All Scripts to create hyperlink for TE Resources 
Updated CLOS for TE resources that had incorrect titling that broke the new functionality


courtrightd (approved):
Approved!

Issue #1928, opened 2024-08-08
ACTIONS - MA-EPD EI FIAT Bug on searching for non M01 


C-Love:
Error reported that the MA-EPD FIAT script does not work when a member other than M01 is selected. In review, this functionality has been created but since it relies on a EMSearch and the only search is of the Member Number, the script often finds the wrong row to use. 

Additional search information should be added to select the right row. Then testing is needed to see if that will resolve the issue.

C-Love:
This appears to be working.

PR #1936, opened 2024-08-14
#1928 ACTIONS - MA EPD EI FIAT - Search for MEMB better

C-Love:
The search process looks only for a number so it was falsely identifying the row a non-01 member was in. 

Added additional characters to the search for it to find the person on ELIG/HC


courtrightd (approved):
Approved!

PR #1939, opened 2024-08-16
1364 NOTES - HC EVAL, require some review

MDR22:
Updated error functionality to: 
- Require review of BILS information if a BILS panel exists
- Require that user enters notes about each type of income (BUSI, UNEA, JOBS) if the income exists. If the income does not exist, it does not require notes to be added


courtrightd (approved):
Looks good. Approved.

Issue #1946, opened 2024-08-22
Bug on CAF Script - Interview date failing for programs that were already approved


C-Love:
The Interview Date is supposed to be filled in based on the Interview script being run for a case. 

This appears to be working correctly when the PROG and REVW dates are automatically selected when the script run. If the worker needs to 'Add SNAP' or other programs, the script is removing the interview date read from the CASE/NOTE. 

Need to be able to reset the date from the Interview Note if it is lost.

C-Love:
I think it is funny that we JUST talked about this. Confirmed it works if the case isn't acted on to make the programs no longer pending. This update will add stability.

C-Love:
This seems to be working.

PR #1949, opened 2024-08-22
#1946 NOTES - CAF - interview date from script run bug

C-Love:
In the instance that a program was not originally selected, the interview date from the Interview script run was getting 'forgotten'.

This 'remembers' the date.


courtrightd (approved):
Approved!

Issue #1948, opened 2024-08-22
Bug on ELIGIBILITY SUMMARY for HC Spenddown Type Determination


C-Love:
Error on a case was reported where the HC span over 6 months had different spenddown types. MOBL pop-up will list all of them and we needed better clarification and determination of which information should be read and which is not correct for the case in the given month. 

Having a case was extremely important to understanding this issue and identifying where this can be resolved.

C-Love:
This appears fixed.

PR #1950, opened 2024-08-22
#1948 NOTES - Elig Summ - HC Spenddown Type Bug

C-Love:
Need better review of the month to find the right spenddown type and match to the correct data.


courtrightd (approved):
Looks good from here!

Issue #425, opened 2021-04-08
Update Docs Received to be more Dashboard like and have more internal workflow


C-Love:
Right now docs received has a checkbox situation that calls functionality in series. 

This should be more comprehensive. 

Additionally, add functionality for Verifs Requested in case there are more verifs that are needed after receiving docs (pull functionality from CAF/Interview to match Verifs Requested).

C-Love:
One thing to be sure to include is DISA updates/information for LTC option. 

Need a bunch of logic/process review on HOW this will work.

MeganGeissler:
Forms to consider adding to Docs Received based on feedback: 
a. Habitability Inspection:  DHS7123
		b. School verification:  DHS2883, D105
		c. Financial Aid Information DHS2646
		d. Request for Voluntary Vendor Payment (Vendor Start/Stop Request) DHS2984
		
		e. Various Forms for the following items: [Acceptable Verification (sharepoint.com)](https://hennepin.sharepoint.com/teams/hs-es-manual/sitepages/Acceptable_Verification.aspx)
			i. Identity forms
			ii. Proof of ID
			iii. Immigration Documents/Status
			iv. Asset Information 
		f. Needed in Home form -- MFIP Medical Form Needed in Home SED-SPMI?? Kari reached out to Hodan- not supposed to be obsolete- they are fixing this.
			a. Obsolete? [Forms_List - MFIP Medical Form - Needed In Home/SED/SPMI (sharepoint.com)](https://hennepin.sharepoint.com/teams/hs-es-manual/Lists/Forms_List/DispForm.aspx?ID=190)
		
		g. Proof of Relationship: NOTES- Proof of Relationship exists. 
This script can be used at every Application and Eligibility review to define the relationships clearly. It can also be used to detail new relationship proofs received or changes in relationships. Using the checkbox on the right, a verification that is new can be indicated.



IAA & ATR forms basically need to casenote that they were received, correct?

PR #1955, opened 2024-08-28
#425 Simple fixes based on feedback

MeganGeissler:
Fixed a few quicks things based on feedback so they don't sit until my return.


courtrightd (approved):
These all look good, approved and merging.

Issue #37, opened 2020-06-10
Find Member in MMIS with recent HC Eligibility


C-Love:
Create a tool to look up any affiliated cases in MAXIS.

Possible add-on, determine if duplicate PMI exists. This may need its own issue and script.

Request from Aaron Gardner-Kocher on 5/7/19

IlseFerris:
Question - should this one be on the power pad?

IlseFerris:
Per the team's conversation, as this will be primarily a UTILITY type of script vs. a NAV script, and will not be placed on the power pad.

IlseFerris:
Create a function that can also be added into Client Contact, etc. Parallel task #390

C-Love:
Changing the scope here to look for active or recently closed cases in MMIS for an individual as opposed to a whole case. 

The script will select an individual from a case and then use PMI, SSN, DOB, and Name in different combinations to find multiple PMIs in MMIS. These PMIs will all be searched for any active HC span or any HC span that closed in the past 4 months and capture that Case Number and PMI for a display of informaiton.

C-Love:
DaNM -Update the powerpad to replace NAV- POLI TEMP with this new script
![image](https://github.com/user-attachments/assets/a19b913a-3ff8-4c63-9452-bbfa7590d37c)

C-Love:
Power pad and redirect files changed on the LAN today to support this script.

C-Love:
Script is done.

PR #1961, opened 2024-08-30
#37 Replace Nav POLI TEMP with Find MEMB in MMIS on Power Pad

C-Love:
These are the GitHub Updates only and will not deploy to workers. 

Once reviewed and approved, I will also directly update the powerpad files and add the redirect to the LAN for the workers to access the new functionality.


courtrightd (approved):
Go for it!

Issue #1958, opened 2024-08-29
ELIGIBILITY SUMMARY - Remedial Care Warning Logic needs Correction


C-Love:
New clarification about policy has been received regarding remedial care deductions for HC for persons in Type 55 or 56 Facilities. Currently the assessment of if this deduction has been allowed is dependent on GRH status for the person. Clarification is that GRH status does not matter for applying the remedial care deduction.

Elig Summ will provide a popup for case that meet certain criteria to remind the worker that a remedial care deduction is appropriate and clarify the use. It is not inhibiting and is intended as a reminder/support.

This is the criteria we should be using:
1. Person is in a Type 55 or 56 facility
2. Health Care is being processed
3. Heath Care has a spenddown

For these criteria, the Remedial Care reminder should trigger. 

In an attempt to check and support this policy, we are also adding a QCR check for these cases - with the email going to BT.

C-Love:
Resolved - no errors reported.

PR #1964, opened 2024-09-05
#1958 NOTES - Elig Summ - HC Remedial Care QCR to BT

C-Love:
How about we stop getting a million emails. 

I have reviewed this functionality for these warning and quality control review emails are now going to BT. 

There was one issue in evaluating Waiver Obligation and I had to add criteria around this being assessed at $0. 

There was also an issue when the worker selected to not enter the HC Case note - the Ex Parte details still operate and they should not. Updated the logic to skip these steps if the worker says 'No - but continue the script' in the Remedial Care warning pop-up.


courtrightd (approved):
I trust you, bro. Approved.

Issue #1848, opened 2024-06-03
NOTES - DISASTER FOOD REPLACEMENT, update to reflect 06/2024 POLI/TEMP guidance


MDR22:
POLI/TEMP updated 06/2024 includes additional instructions around entering a denial reason and/or sending a SPEC/MEMO for denied requests. Update script to reflect new instructions.

PR #1972, opened 2024-09-10
#1848 NOTES - DISASTER FOOD REPLACEMENT, update to reflect 06/2024 POLI/TEMP guidance

MDR22:
POLI/TEMP updated 06/2024 includes additional instructions around entering a denial reason and/or sending a SPEC/MEMO for denied requests. Updated script to for situations where request is denied due to an affidavit not being received or if the loss was reported late to direct worker to follow TE02.11.07 (Replacing Benefits). In situations where request is denied for other reasons, added script functionality to generate a SPEC/MEMO.


courtrightd (approved):
Approved!

Issue #1974, opened 2024-09-12
ADMIN - DAIL FMED DEDUCTION, improve handling for DAIL to prevent error at SELF


MDR22:
Received report of script getting stuck at SELF screen, appears to be an issue with do loop for DAIL that results in script getting kicked back to DAIL and then erroring.

PR #1978, opened 2024-09-13
#1974 ADMIN - DAIL FMED DEDUCTION, improve handling for DAIL to prevent error at SELF 

MDR22:
Added exit condition if "NO MESSAGES" error appears after deleting a DAIL, which can cause an error where the script gets stuck at SELF.


courtrightd (approved):
Approved!

Issue #1967, opened 2024-09-06
On Demand Dashboard bug in saving strings with an apostrophe into SQL


C-Love:
Reported from QI worker that the script errored when trying to save information about a case that was on hold. The syntax of the error indicated this was due to an apostrophe included in the string, which is a reserved character in SQL.

There is simple handling to remove any apostrophes from strings (usually in notes) before entering into SQL.

C-Love:
No errors created from this update - should be working well.

PR #1983, opened 2024-09-17
#1967 ADMIN - On Demand Dashboard - remove single quote from strings

C-Love:
When information is entered into SQL from the Dashboard - there is only one field were workers enter text free form - the tracking notes. Other string SQL fields are managed by text hard coded in the script. 

Testing ability is limited for this script as it functions on live cases from the On Demand list. No syntax errors were found. JH is scheduled for On Demand this week, I have sent her an email of the changes and to let us know any issues that arise this week.


courtrightd (approved):
Looks like it will work.

Issue #1940, opened 2024-08-19
HC Evaluation - reschedule ex parte cases error handling for 15th or later


courtrightd:
When rescheduling cases that fail ex parte during phase 2, if the processing occurs on or after the 15th of the month, paper renewal paperwork will not go out to the resident. These cases must be pushed out an extra month. Add error handling to make sure the correct reschedule month is chosen when processing phase 2 late.

PR #1984, opened 2024-09-17
#1940 Ex Parte Reschedule correct month if date is after 15th

courtrightd:
Sets the default renewal month to CM+3 when the date is after the 15th and user chooses to reschedule ex parte.
Fixes STAT/REVW coding instructions for AR dates.


C-Love (approved):
I trust you bro.

Solid - should work perfectly.

Issue #1613, opened 2023-12-19
ACTIONS - Earned Income Budgeting - allow update/view the semi-monthly check dates


C-Love:
Report received that the script did not correctly determine the days of the month that a person was paid when the pay frequency of semi-monthly was received. 

The two aspects we should address with this issue is:

1. Allow users to view and change the selected days for the semi-monthly frequency option from the confirmation dialog. 
2. Have the script shift to non-weekend days when creating the pay dates.

Report from AGK:
Information: This client appears to be paid the 15th and last of the month, or the weekday before those dates when they're weekends. I think semi-monthly pay especially has lots of quirks like that and the script could use more functionality to adjust the "normal" pay cycle. In this case it seemed to guess that the 9/29 check actually means usually paid the last day of the month, but kept the 13th as the other expected pay date. Not sure how worth it making changes will be though as long as the frequency is correct. It could just make things too complicated to worry too much about getting the exact expected dates right.

C-Love:
Done - the script isn't broken - should be working.

PR #1985, opened 2024-09-17
#1613 ACTIONS - EIB - update SemiMonthly dates handling

C-Love:
These updates allow workers to specifically identify the SemiMonthly pay dates because the script sometimes gets the determination wrong. It would be hard to make the script perfect but allowing more visibility and an override will make the script work more often. 

Updates:
- Create a function to display and update the SemiMonthly pay dates so it can be easily called. 
- Add a display of the semi-monthly pay dates to the confirmation dialog.
- Add a button to call the SemiMonthly dates dialog.

There were also a few bugs discovered while fixing the issue.


courtrightd (commented):
I tested. It worked. Approved.

Issue #1938, opened 2024-08-16
CAF - allow to proceed with missing EMER interview


courtrightd:
To align with current business process, we are going to allow the script to proceed if there is a blank interview date for emergency, but other pending programs have an interview date correctly coded.
Script should not proceed if emer is only pending program, the case is not at review, and there is not an interview date.

PR #1986, opened 2024-09-17
#1938 NOTES - CAF, allow script to proceed even if EMER interview date is missing

MDR22:
Updated script so that it will proceed even if EMER interview date is missing as long as there are other pending programs.


courtrightd (approved):
Approved!

PR #1989, opened 2024-09-19
425 documents received updated dashboard feel and new forms

MeganGeissler:
- Revised DIET functionality to only update DIET in MAXIS with the highest $$ diet of those that overlap 
- Updated Shetler form to capture form details, replaced checkbox to update ADDR, HEST, SHEL with buttons as well


MDR22 (approved):
Tested several times and didn't run into any issues. Looks good to me!

Issue #1971, opened 2024-09-10
ELIGIBILITY SUMMARY - Collect Data on MSA cases with UNEA 44 or Shelter Needy


C-Love:
Per request from KW from QI and Appeals there have been appeals and accuracy in processing MSA on cases that have no SSI or have Shelter Needy. 

There is a need to asses how many of these cases are processed in error. In order to get an idea of the scope, there is a request from KW to add a QCR for these cases for manual review. TP from QI should be able to confirm the need for this project. 

Updates Needed:
1. Add a QCR for MSA with Shelter Needy
2. Add a QCR for MSA with UNEA coded 44
3. QCR reports will only be set to text cookies.
4. Create a Report Out script for cookie based QCRs to be used by BZST.

C-Love:
This is deployed and working.

PR #1990, opened 2024-09-19
#1971 New Script - ADMIN - Review QCR Reports 

C-Love:
Script allows the information recorded as QCR reports to be compiled into an Excel file.


courtrightd (approved):
Tested fine, this looks good!

Issue #1965, opened 2024-09-05
Add a Remedial Care reminder in HC Eval 


C-Love:
If a resident is in a Type 55 or 56 facility, add a reminder text to HC Evaluation to check for the need for Remedial Care amount in BILS to support accurate approvals. 

This should not stop the script from operating as normal, just provide information and support to the workers.

There is already more supportive functionality for this policy in Eligibility Summary.

C-Love:
The script didn't break. Updates completed.

PR #1991, opened 2024-09-20
#1965 NOTES - HC Eval - Add Remedial Care Reminder

C-Love:
Confirmed with BT that the phrasing and functionality will be supportive.


courtrightd (approved):
Approved!

PR #1992, opened 2024-09-20
#1971 NOTES - Elig Summ - bug in reading and eval MSA

C-Love:
In reviewing some of the QCR information it was found that the MSA budget is reading it incorrectly.


courtrightd (approved):
Well, there's your problem lady. APPROVED!

Issue #1994, opened 2024-09-23
ADMIN - Resolve HC EOMC in MMIS - bug due to NAT error


C-Love:
In running the report - the script failed because one worker number threw a NAT error. 

Added handling to identify this error and continue the script run to complete the report.

C-Love:
Done - now we have this nonsense written down somewhere

PR #1995, opened 2024-09-23
#1994 ADMIN - Resolve HC EOMC in MMIS - bug with NAT error

C-Love:
I don't know that this will be an ongoing issue but since a similar result happened on another script, I was thinking we could save this to record the issue/solution and maybe remember it.


courtrightd (approved):
Approved I guess.

Issue #1108, opened 2023-01-09
ELIGIBILITY SUMMARY - add support for Significant Change


C-Love:
Update functionality to support processing significant change.

Processes to address:
- FIAT
- WCOMs
- Possible MEMOs
- Possible MONY/CHCK

POLI TEMP TE02.13.11

This is some of the work outlined in #94

C-Love:
This appears resolved. No issues reported.

PR #1996, opened 2024-09-24
#1108 NOTES - Elig Summ - Sig Change CASE NOTE

C-Love:
Read ELIG/MFSC for information and create a Significant Change specific NOTE


courtrightd (approved):
A-Ok from here.

Issue #1998, opened 2024-09-25
function start_a_blank_CASE_NOTE() - improve error handling


MDR22:
Add additional error handling to capture more details in script lowdown such as the panel where the error occurs and capturing the MAXIS mode, even if not on CASE/NOTE

PR #2000, opened 2024-09-25
#1998 function start_a_blank_CASE_NOTE() - improve error handling 

MDR22:
Add additional error handling to capture more details in script lowdown such as the panel where the error occurs and capturing the MAXIS mode, even if not on CASE/NOTE


C-Love:
Just a thought - would it be helpful to capture row 2 - to grab what screen we are on?

MDR22:
Updated to read the panel name too

courtrightd (approved):
Approved!

Issue #1376, opened 2023-06-30
IV-E Script Updates


IlseFerris:
7 scripts need updating and new one created for Adoption Assistance, IVE, MA Needy Child, and Trial Home Visit. 
Project request submitted by HSS, BS. Saved request in LAN drive. 

Scripts to Update:

- [x] Trial Home Visit
- [x] Adoption Assistance
- [x] IVE
- [x] Elig Review
- [x] MA Needy Child
- [x] Northstar Kinship - New Script

C-Love:
Discussed with IVE Team - test for 2 weeks unless issues or changes are needed.

C-Love:
Script changes confirmed by the team, completed.

PR #2001, opened 2024-09-26
#1376 IVE - Adoption Assistance - switch close and placement ended

C-Love:
Report from IV-E Workers that the details of these two options should be aligned in this way.


courtrightd (approved):
Approved

Issue #2002, opened 2024-09-27
Application Received BUG for Cash pending when CASH is active


C-Love:
Error reported for a case that is pending a cash program but cash is already active.

Need to review pending in REPT/PND2 instead of just CASE/CURR.

C-Love:
Fixed!

PR #2003, opened 2024-09-27
#2002 NOTES - App Recvd - second cash pending bug fix

C-Love:
If Pending is not found on CASE/CURR - we need to check REPT/PND2 before cancelling out of the script.


courtrightd (approved):
Looks good! Re-worded changelog, approved from here if you are fine with my edits.

C-Love:
Verbiage changes look good to me.

PR #2008, opened 2024-09-30
#425 SHEL update tweaks for Docs Recvd

C-Love:
Update to the SHEL functions for display and update now that we are actually using it. 

The format of the Y/N wasn't aligned with the droplists and there was some extra entry that happened. 

Additionally, we can now turn off the button for the totals. This instance should be set up to start with a member instance of SHEL and not defaulted to starting on Totals.


MeganGeissler (approved):
Tested and looks good to me

courtrightd (approved):
Looks good, works good! Send it!

Issue #1896, opened 2024-07-03
Retire NOTES Change Report Form Received


MeganGeissler:
Retire this individual script once #425 #1298 Docs Received is published. 

- [x] Move scripts files to retired scripts folder
- [x]  Updated main menu
- [x]  Update CLOS with retirement dates for each script.
- [x]  Remove script from instructions library on [SharePoint](https://hennepin.sharepoint.com/teams/hs-economic-supports-hub/BlueZone_Script_Instructions/Forms/AllItems.aspx)
- [x]  Confirm if HSR manual needs to be updated

PR #2011, opened 2024-09-30
#1895 #1896 Retired Change report form received and ltc hospice form …

MeganGeissler:
Retired Change report form and ltc hospice form: moved scripts to retired folder, updated main menu, updated CLOS


courtrightd (approved):
Send it!

Issue #388, opened 2021-03-15
Health Care - Create CASE:NOTE for Auto Close Cases with closure detail (REVW MONT Closure)


C-Love:
Take the old REVW MONT Closure process and reevaluate and merge into our processes.

The idea here is to capture some details FROM MAXIS on cases that will/have autoclosed due to REVW and add it to the CASE:NOTE for quick case understanding.

Effort in Quality CASE NOTES.

This issue is for any case associated with any 6 mo or 12 mo REVW that is not pulled in On Demand- so any review or program that would NOT end up on the REVIEW REPORT.

There are separate issues for:

- All REVIEW REPORT REVW Cases - #387 
- ALL HRFS - #389

IlseFerris:
Changed scope of this issue to Health Cases when HC renewals are back in effect. All other programs on REVW are already accounted for in #387.

C-Love:
The current PHE is set to expire after July 14 2022. If this is not extended HC renewals COULD resume for October 2022. This could be later and if the PHE is extended, it would likely be extended to Mid-October which would mean the renewals would be months after that time. It appears the earliest renewals could happen would be October. Recommend to move this to Q3 as we should know by that time if this work would be needed for cases to be processed in Sept for Oct renewals.

C-Love:
ADDITIONALLY - review functionality in Review Report to ensure that all required steps are in place for HC - tracking, notices, statistics, etc.

C-Love:
Discuss at next meeting - Does the changes to HC review processing change the timing on this issue?

C-Love:
This will not need to be run until 2/1/2024 at the very earliest, rescheduling this for Q1 2024 to be completed in January.

C-Love:
Done!

PR #2012, opened 2024-10-01
#388 Review Report - Add Autoclose CASE NOTE for HC

C-Love:
Include HC in the End of Processing Month script runs


courtrightd (approved):
Looks good from here!

PR #2021, opened 2024-10-02
1781 housing support dhs rate updates

IlseFerris:
Updates have been sent to DHS via text file for testing. Rates changing for 2024 is the only update.


IlseFerris:
@courtrightd - DHS user just confirmed this is working as intended.

Issue #2024, opened 2024-10-03
ELIGIBILITY SUMMARY - bug in reading ELIG HC


C-Love:
Error reported when reading ELIG HC for cases with INEL spans in HC. 

The Assets pop-up won't open and an inhibiting error appears if the HH Count has not completed. 

Currently, we don't read any asset information in ELIG - so we can just not open this pop up.

C-Love:
This is fixed.

PR #2025, opened 2024-10-03
#2024 NOTES - Elig Summ - don't open useless pop-ups

C-Love:
Bug fix for HC cases when the Asset window tries to open but the HH Count has not been completed. We don't need to open this pop-up anyway.


courtrightd (approved):
Makes sense, approved!

Issue #2036, opened 2024-10-09
ADMIN - DAIL DECIMATOR, incorporate DAIL HIRE message deletion functionality


MDR22:
Add functionality to DAIL to delete HIRE messages (SDNH and NDNH) that are over 12 months old

PR #2042, opened 2024-10-15
#2036 ADMIN - DAIL DECIMATOR, add functionality to delete HIRE messages over 12 months old

MDR22:
Added functionality to delete both SDNH and NDNH HIRE messages that are over 12 months old from the current month. I did not run the script on these x numbers so they would be good options for testing: EX3, EX4, and EX5. Otherwise, I ran it on all other x numbers.


courtrightd (approved):
Added a commit to fix a syntax error / missing end if. Please review my commit to make sure  it is the only changes that were needed. Tested on EX3, worked fine. Did not clear the other two case numbers. 
Approved once you double check my work. :-)

MDR22:
Good catch! I tested and it works.

Issue #1782, opened 2024-04-14
NOTES - SNAP Waived Interview Sundown


IlseFerris:
Sundowning Postponed: SNAP Waived Interview Extended. Documentation includes drafted code and survey.
T\Eligibility Support\Restricted\QI - Quality Improvement\BZ scripts project\Projects\SNAP Waived Interview\ Sundowning SNAP Waived Interview

Tasks include:
Retire NOTES - SNAP WAIVED INTERVIEW 
- [x] Move scripts files to retired scripts folder
- [x] Updated main menu
- [x] Update CLOS with retirement dates for each script.
- [x] Remove script from instructions library on [SharePoint](https://hennepin.sharepoint.com/teams/hs-economic-supports-hub/BlueZone_Script_Instructions/Forms/AllItems.aspx)

Other scripts to update
- [x] Update MISC - DESERT ISLAND's Dialog to remove NOTES - SNAP WAIVED INTERVIEW script - Align with retirement of waived interview.
- [x] Update Interview - Remove follow up contact if it exists
- [x] Update NOTES - Client Contact: Remove 'follow up contact' Lines 122-203. Align with retirement of waived interview?
- [ ] Remove snap waived interview verbiage from: [NOTES - CLIENT CONTACT.docx (sharepoint.com)]

### Associated Issues 
- #689: SNAP Work Rules will be added to the Interview script
- #1753: Review Report allow SNAP appointment notices and NOMI's

MeganGeissler:
SNAP Waived interview was extended until 10/31/2024. Sundowning this project is pushed out. 

Food and Nutrition Services (FNS) has offered the option to request an extension of the SNAP Interview Waiver through October 31, 2024. FNS is approving this six-month extension under the extraordinary circumstances criterion due to challenges associated with unwinding from the pandemic. FNS does not plan to approve additional extensions to this waiver after October 31, 2024.

The Department of Human Services SNAP Team has decided to accept this offer and submit the request to extend the SNAP Interview Waiver through October 31, 2024. The waiver extension will only be available to counties and Tribal Nations who implemented the waiver during May 2023 – April 2024.

PR #2053, opened 2024-10-22
#1782 Added msgbox reminding users waiver is ending

MeganGeissler:
Adding msgbox reminding waiver is ending.

Will retire the script using the same branch next week.


courtrightd (approved):
Approved, thank you!

Issue #2056, opened 2024-10-24
NOTES - HC Evaluation - Bug BILS dialog maxing out due to large number of BILS pages


MeganGeissler:
Script reads all bills listed on BILS panel, in this example there are so many bills it maxes out the script. 

Need to set a limit on how much it reads. Probably pick whatever fits into the dialog nicely.

PR #2057, opened 2024-10-24
#2056 BILS dialog bug fix

MeganGeissler:
Added a limit of reading 2 panels worth of BILS so that it does not exceed the dialog.


courtrightd (approved):
Approval bestowed.

PR #2058, opened 2024-10-25
#1753 ADMIN - Review Report - Restart SNAP Appt Notices

C-Love:
Update Review Report to send notices for all cases with an interview requirements as the interview waiver has ended.

Ran through Appointment Notices today with no issues.

Additionally added a a call to save the workbook after the Appointment Notice and NOMI runs because I have had some issues with those not being saved at the end of the script run.


courtrightd (approved):
I hereby bestow my approval upon thee.

MeganGeissler (approved):

Issue #2061, opened 2024-10-28
Eligibility Summary - Handle for non-number entries


C-Love:
Report from worker that the amounts on SHEL were entered as '?' and so the script errored. While this is not typically a suggested process, SHEL allows for this entry and we need to accommodate it in the script run to make this a 0 for the calculations.

C-Love:
Issue completed

PR #2062, opened 2024-10-28
#2061 - Elig Summ - SHEL Question Marks are 0 amount

C-Love:
Add handling to turn '?' entries to become 0 for calculations


courtrightd (approved):
Looks like a good solution to me, thank you!

PR #2063, opened 2024-11-01
1782 sundown snap waived remove access to script

MeganGeissler:
First step in sundowning snap waived interview - plan to pull in Monday morning upon approval
-Added retirement date to CLOS so script doesn't appear for end users to select 
-Removed msgbox in SNAP waived interview notifying users that the script is ending (since you'd only hit this through the return contact)
-Added retirement message to main menu


courtrightd (approved):
Approved.

Issue #2064, opened 2024-11-05
function start_a_blank_case_note - ensure new case note is opened


MDR22:
User reported issue with script reporting that case note successfully completed despite script not actually creating new case note. This is an issue if the user starts the script when already on an open case note from a previous day. Add functionality to ensure that a NEW case note is open, even if script is started on a case note from the previous day.

PR #2065, opened 2024-11-05
#2064 function start_a_blank_case_note - Check if case note already open and open a new case note if needed

MDR22:
Add functionality to start_a_blank_case_note to check if already on new or existing case note. If already on case note then it PF3s and then PF9s to open a blank case note. User had reported issues with scripts not correctly creating a case note when they were already in a case note.


MeganGeissler (approved):

courtrightd (approved):
Approved, thank you!

Issue #2055, opened 2024-10-23
NOTES - Interview edrs_notes_for_case variable not casenoted anywhere


MeganGeissler:
EDRs Case Notes (edrs_notes_for_case) entry is not called in case notes anywhere, need to add this.

Issue #1539, opened 2023-10-18
Review Interview script and update DropListBoxes to ComboBoxes where it makes sense


C-Love:
If a person is not in MAXIS before the interview script is started, they won't be in the list of options in any droplistbox. We can change these to ComboBoxes to allow for manual entry of people that are reported in the interview or aren't in MAXIS for whatever reason.

IlseFerris:
This should be worked on at the same time as #560 so this will also travel to another quarter.

MeganGeissler:
INTV/CAF 1: who are we completing the interview with - COMBO
CAF MEMBs: Add Person uses new_clt and sets selected_memb = new_clt to determine new count for array and to store info 
Q7-11: Principal wage earner - DROP BOX
CAF QUAL Q: Qualifying questions HH member - COMBO
CAF Last Page: Signatures of primary/other - COMBO

I think we are covered w/ combo boxes at this point.

Issue #1198, opened 2023-03-08
NOTES - INTERVIEW: Add Housing Supports Questions/Supports


IlseFerris:
We are having some issues with HSR's attempting to conduct an interview that does not incorporate all programs. To help reduce this interview questions specific to the Housing Support program will be added. There may also be some workflow/communication steps.

Very similar to issue #892.

IlseFerris:
Suggested questions from DHS Housing Support staff: 

> "Are you working with a licensed Housing Support provider"?
> If yes - "who is the provider"? 
> If no - no elig
> IDK - pend application and send verification requests
> Explain to resident that they would have to have been referred to a licensed Housing Support provider through Coordinated Entry Services and have a landlord who is willing to accept this type of funding.

Issue #603, opened 2021-09-23
NOTES - Interview - Add Update of HEST Panel


C-Love:
Update the handling of HEST information in the Interview script - #62, then have the script UPDATE HEST with information gathered in the interview.

PR #2066, opened 2024-11-06
568 1198 1539 2055 603 notes interview revisions

MeganGeissler:
-Reviewed dialogs for combox opportunities #1539
-Script updates ADDR if Update Information is selected #568
-Added housing support questions to ADDR dialog #1198
-Script updates HEST if persons paying is identified and heat/ac, electric, or phone is noted. Need to update HEST function once FuncLib is available to ensure we are clearing fields that do not have an entry.  #603
image
-Bug: Added edrs_ntoes_for_case to case notes #2055

### DIFF
--- a/notes/interview.vbs
+++ b/notes/interview.vbs
@@ -5335,7 +5335,7 @@ function jobs_details_dlg(this_jobs)
 
 		Dialog1 = ""
 		BeginDialog Dialog1, 0, 0, 321, 165, "Add Job"
-		  DropListBox 10, 35, 135, 45, pick_a_client+chr(9)+"", JOBS_ARRAY(jobs_employee_name, this_jobs)
+		  ComboBox 10, 35, 135, 45, all_the_clients, JOBS_ARRAY(jobs_employee_name, this_jobs)
 		  EditBox 150, 35, 60, 15, JOBS_ARRAY(jobs_hourly_wage, this_jobs)
 		  EditBox 215, 35, 100, 15, JOBS_ARRAY(jobs_gross_monthly_earnings, this_jobs)
 		  EditBox 10, 65, 305, 15, JOBS_ARRAY(jobs_employer_name, this_jobs)
--- a/notes/interview.vbs
+++ b/notes/interview.vbs
@@ -9394,18 +9394,9 @@ Call check_for_MAXIS(False)
 If need_to_update_addr = "True" then need_to_update_addr = True
 If need_to_update_addr = "False" then need_to_update_addr = False
 
+
 If need_to_update_addr = "True" Then 
-	If right("0" & DatePart("m", address_change_date), 2) <> MAXIS_footer_month OR right(year(address_change_date), 2) <> MAXIS_footer_year Then 'If address change date doesn't match maxis footer month/year must update maxis footer month/year for ADDR udpate
-		orig_footer_month = MAXIS_footer_month
-		orig_footer_year = MAXIS_footer_year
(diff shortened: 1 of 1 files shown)


notes/interview.vbs
```diff
@@ -1194,6 +1194,11 @@ function define_main_dialog()
 			col_3_1 = 335
 			col_3_2 = 380
 
+			EditBox 225, y_pos -5, 145, 15, all_persons_paying
+			Text 	20, y_pos , 210, 10, "List household members who pay utilities (separated by commas):"
```
MDR22, line 1198: Minor edit but the text field slightly overlaps the entry field

image

notes/interview.vbs
```diff
@@ -1194,6 +1194,11 @@ function define_main_dialog()
 			col_3_1 = 335
 			col_3_2 = 380
 
+			EditBox 225, y_pos -5, 145, 15, all_persons_paying
+			Text 	20, y_pos , 210, 10, "List household members who pay utilities (separated by commas):" 
+			y_pos = y_pos + 15
+
```
MDR22, line 1200: I wonder if this should be a dropdown or checkbox of some sort. Since it is a free text field, the script will error if the incorrect HH memb numbers are used (such as putting in the names of the HH members instead of the number). Alternatively, maybe the instructions could be clearer that the HH Memb Number needs to be used - "List the household member numbers who pay utilities (separated by commas):"

notes/interview.vbs
```diff
@@ -9329,6 +9391,55 @@ Loop until are_we_passworded_out = FALSE
...
+		prosp_electric_yn = " " 
+		prosp_electric_units = "  "
+	End If 
+
+	If question_15_phone_yn = "Yes" Then 
+		hest_update_phone = TRUE
+		prosp_phone_yn = "Y"
+	Else
+		prosp_phone_yn = " "
+		prosp_phone_units = "  "
```
courtrightd, line 9433: Looks like the function doesn't blank out the number of units, which errors the panel. What does work is saying "N" and leaving the "01" on the panel when updating - so changing these to
 Else
Prosp_phone_yn = "N"
should actually be the working solution

notes/interview.vbs
```diff
@@ -9067,7 +9124,12 @@ If vars_filled = FALSE AND no_case_number_checkbox = unchecked Then
...
+    Call access_ADDR_panel("READ", notes_on_address, resi_line_one, resi_line_two, resi_addr_street_full, resi_addr_city, resi_addr_state, resi_addr_zip, resi_addr_county, addr_verif, homeless_yn, reservation_yn, living_situation, reservation_name, mail_line_one, mail_line_two, mail_addr_street_full, mail_addr_city, mail_addr_state, mail_addr_zip, address_change_date, addr_future_date, phone_one_number, phone_two_number, phone_three_number, phone_one_type, phone_two_type, phone_three_type, text_yn_one, text_yn_two, text_yn_three, addr_email, verif_received, original_information, update_attempted)
+
+	resi_line_one = ""
```
courtrightd, line 9129: If the update date doesn't match the footer month, MAXIS errors. We need to find a way to match dates to handle for this.

courtrightd (changes requested):
I found two error situations with the panel updating - see comments.
All else looks good!

notes/interview.vbs
MeganGeissler: @courtrightd I removed the handling for this and will handle for it in the func lib once it's available based on discussions this morning.

courtrightd (approved):
New updates are working, approved!

PR #2071, opened 2024-11-13
#2064 NOTES - Elig Summ - Back out of all Case Notes after completed

C-Love:
Update to Eligibility Summary to ensure it will work well with the updates to start_a_blank_case_note. 

When a CASE/NOTE is created and all information is entered, we need to back out by pressing PF3 before continuing the script.


MDR22 (approved):
Looks good!

Issue #2072, opened 2024-11-13
On Demand - Automate the handling of data folders by month


C-Love:
The On Demand script requires a month specific folder to store daily action data. This was being completed manually but it is fully able to be automated. 

There needs to be a folder for the current month and any folder that is from two months prior should be moved to the archive.

PR #2073, opened 2024-11-13
#2072 On Demand Applications - automate folder management

C-Love:
Folder created if missing and moved if too old. 

No rush on this review, November is correctly set up, so it won't be particularly relevant until December.

I ran this to set up November correctly.


courtrightd (approved):
Looks good!

Issue #2070, opened 2024-11-13
NOTES Doc Received - Bug Case Noting 


MeganGeissler:
Eliminate unnecessary use of start a blank case note

PR #2074, opened 2024-11-13
#2070 fixed bug related to start blank case note

MeganGeissler:
fixed usage of blank case note in doc received on branch with func lib updates


MeganGeissler:
@courtrightd - Isn't it weird we don't see mark's changes to the func lib as a file? I do see if I click on the branch link up top and look at that file...

courtrightd (approved):
Looks good, lets hope it merges ok!

Issue #568, opened 2021-08-18
NOTES - Interview - Add Update of ADDR Panel


C-Love:
Update the functionality to update ADDR during the script run. 

LOW PRIORITY FOR Q3 - MAY MOVE TO Q4

PR #2076, opened 2024-11-15
2075 update addr & hest

MeganGeissler:
Updated ADDR: Updated handling of which month to update ADDR in and maintain variables after. Connected to #568.

Updated HEST: connected to issue #603. Updated HEST function to clear/blank out fields in case there is already info on the panel for that field. 

Updated Doc Rec: Impacted by ADDR update, no longer needed handing within doc rec for which month to update ADDR in.


courtrightd (approved):
Functions perfect, approved!

PR #2079, opened 2024-11-20
#560 NOTES - Interview - handling for different application forms

C-Love:
Incorporates the script file that builds a form based on a class object kept in a different script file. 

This should change the interview experience based on the user selection during the script run.

### DIFF
--- a/misc/interview-forms-classes.vbs
+++ b/misc/interview-forms-classes.vbs
@@ -2215,6 +2215,28 @@ class form_questions
 		Set info_number = xmlDoc.createTextNode(dialog_height)
 		XML_dialog_height.appendChild info_number
 	end sub
+
+	sub add_to_SRL()
+		script_run_lowdown = script_run_lowdown & vbCr & number & ". " & dialog_phrasing
+		script_run_lowdown = script_run_lowdown & vbCr & "info_type - " & info_type
+		script_run_lowdown = script_run_lowdown & vbCr & "caf_answer - " & caf_answer
+		If sub_number <> "" Then
+			script_run_lowdown = script_run_lowdown & vbCr & sub_number & ". " & sub_phrase
+			script_run_lowdown = script_run_lowdown & vbCr & "sub_answer - " & sub_answer
+		End If
+		script_run_lowdown = script_run_lowdown & vbCr & "write_in_info - " & write_in_info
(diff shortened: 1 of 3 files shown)


MeganGeissler:
If I run the HUF cancel out and use the save my work feature to restart, using the 'enter' button to switch tabs does work. Instead of switching tabs it keeps giving me question 2 as a dialog popup. 
image

C-Love:
Ok - I have a fix for it. Looks like the button definition restore had an error.

MeganGeissler:
@MDR22  - I second the save your work -- i was thinking adding the form type to the dialog header once you get into questions could help clarify for the scenario you described (to some extent).

C-Love:
Good catch on the navigation, I will look at it. 

The summary stuff didn't change at all, so I am not sure what is happening there. It should leave the script before it hits any of the changes. I will see what's happening with all of that. 

I thought about this with the save your work. How much do we think that will be a reasonable action? I mean, if they pick the wrong form, they should restart and not load the saved information. We would have to restructure the dialogs to fix this. I am leaning toward this is a user error we shouldn't account for.

C-Love:
We can have the form listed on the main dialog easily but knowing that there is a discrepancy would cause some issues if they opt to not save your work.

MDR22:
> We can have the form listed on the main dialog easily but knowing that there is a discrepancy would cause some issues if they opt to not save your work.

I think that's fair. It wouldn't make a lot of sense if they selected the wrong form upon reloading the save your work. I think some language could be added just to indicate which form they selected previously so they can continue. But I also think it is a user error that probably won't come up?

C-Love:
It may come up, I think we can just reasonably tell them not to restore wrong selections.

We could add phrasing to the save your work question message that the form selection will be overwritten. We just can't know what the previous selection was until we load the save your work detail.

C-Love:
Restore message box:
image


Dialog Title:
image

C-Love:
I figured out what the navigation issue was @MDR22  - the script was trying to go to a page that didn't exist on the scenario you ran. Fixing it caused an issue with the error message displays. So, I think I have that all ready to go.

courtrightd (approved):
Looks like we did some good testing and updates, let's send this thing. Approved!

Issue #2075, opened 2024-11-15
FuncLib - Update ADDR and HEST


MeganGeissler:
ADDR: Determine month to update ADDR panel based on addr_effective_date
- If addr_effective_date is after the footer month/year from the initial dialog, then we need to navigate to the addr_effective_date month and make updates there. Will need to make sure we don't use variables that will impact other updates. 

HEST: Interview script updates HEST if persons paying is identified and heat/ac, electric, or phone is noted. Need to update HEST function to ensure we are clearing the unit fields using "  " for utilities that are not a Y. Do this for retro and prosp.
![image](https://github.com/user-attachments/assets/8493cead-fde3-4f7b-a94b-0126a533e830)

PR #2081, opened 2024-11-20
#2075 HEST function - handled for editing/creating new HEST panel

MeganGeissler:
User reported issue with interview script getting stuck on HEST. Moved if statement criteria up so that it won't create/edit a panel unless it has identified a person paying


courtrightd (approved):
Approval granted!

Issue #2094, opened 2024-11-29
Eligibility Summary - ELIG HC updates


courtrightd:
Update ELIG Summary to read ELIG/HC with new coordinates effective Dec. 1.

PR #2098, opened 2024-12-02
#2094 Eligibility Summary BSUM coordinate updates

courtrightd:
This has briefly been tested - could use another test case or two. Make sure to test with footer month set to both 12/24 and 1/25, as the footer month when entering ELIG is what changes screen placement.


MeganGeissler (approved):
Tested a few times and was able to successfully run through the whole script!

Issue #2087, opened 2024-11-26
HC ELIG Panel Updates - ACTIONS - LTC Spousal Allocation FIATER


MeganGeissler:
HC ELIG (BSUM) Panel coordinates are changing. Update as impacted. 
![image](https://github.com/user-attachments/assets/7bc31857-91e9-44ad-a3bb-c456b41c9026)

PR #2100, opened 2024-12-02
#2087 updated actions ltc spousal fiater to align with panel update

MeganGeissler:
Updated to include if statement to handle for panel update


courtrightd (approved):
Tested fine, approved!

Issue #2088, opened 2024-11-26
HC ELIG Panel Updates - BULK - Find maepd medi cei 


MeganGeissler:
HC ELIG (BSUM) Panel coordinates are changing. Update as impacted. 

![image](https://github.com/user-attachments/assets/3f7c4552-4946-46d1-837a-ccc844d21522)

![image](https://github.com/user-attachments/assets/5ebf102d-c762-4b58-bc58-f0472c324114)

PR #2101, opened 2024-12-02
#2088 updated to align with panel updates

MeganGeissler:
Updated with if statement to align with panel updates


courtrightd (approved):
Looks good, approved!

Issue #2089, opened 2024-11-26
HC ELIG Panel Updates - NOTICES - method b wcom


MeganGeissler:
HC ELIG (BSUM) Panel coordinates are changing. Update as impacted. 

![image](https://github.com/user-attachments/assets/5a2e9685-885e-4fc1-b478-e00cfafd9f08)

PR #2102, opened 2024-12-02
#2089 updated method b wcom to align with panel updates

MeganGeissler:
Updated with if statement to align with panel updates.


courtrightd (approved):
Looks good, approved!

Issue #2091, opened 2024-11-26
HC ELIG Panel Updates - ADMIN - Case Transfer


MeganGeissler:
HC ELIG (BSUM) Panel coordinates are changing. Update as impacted.

![Image](https://github.com/user-attachments/assets/dd889158-6bd6-40a0-b932-0529acfedbd1)

PR #2104, opened 2024-12-02
#2091 updated bulk case transfer to align with panel updates

MeganGeissler:
Updated with EMSearch to find the correct row and read the correct position based on any month/year. 

I added logic if row and col = 0  then var = "" since it was holding the previous variable value. Let me know if this should be something different.


MDR22 (approved):
Didn't run into any issues testing. Looks good!

Issue #2107, opened 2024-12-03
HC Eval Error with looping forever


C-Love:
Issue reported where the script is pulling an old footer month and then looping forever. 

It appears the dates are not being viewed as dates by the script and causing all sorts of issues.

C-Love:
This appears to be resolved.

PR #2108, opened 2024-12-03
#2107 - NOTES - HC Eval - force dates to be dates

C-Love:
Make a date using the footer month and year and force it to be a date. This should create the right footer month information.


MDR22 (approved):
Tested and fixes looping issue. Looks good!

C-Love:
Since this is an inhibiting error for cases that hit this issue, I am pulling in with the one review. Thanks @MDR22

Issue #2038, opened 2024-10-10
DEU - MATCH CLEARED: Add functionality to capture Claim # required for IULB screen 


MeganGeissler:
This was a bug/spur off of #1764. Need to add functionality to capture Claim# from end user to enter into MAXIS for CC overpayment and maybe overpayment now optional as well (?). This is a bug/request from Autum Obrien.

Case examples: 699715 has 2 program FS and MFIP with overpayments on the wage match. 

Issue Notes from IF in OneNotes/Discovery Notes/DEU

MeganGeissler:
AO confirmed it's working correctly

PR #2119, opened 2024-12-10
#2038 Add claim number capture to DEU - Match Cleared 

MeganGeissler:
Added a dialog to capture claim numbers for CC - this should allow the script to enter  resolution_status, claim number(s), and eliminate dail message (I believe this happens automatically when the match is cleared) 

DEU team is in the middle of clearing the new quarter’s matches so they can test it for us.


courtrightd (approved):
Approval bestowed.

PR #2123, opened 2024-12-10
2122 abawd eval date bug

IlseFerris:
This is the same date functionality used in ACTIONS - TLR REPORT. I have live cases if you want to use them to test.


courtrightd (approved):
This looks good, approval granted!

Issue #1952, opened 2024-08-23
Add QC Review to ELIGIBLITY SUMMARY for Supportive Housing Disregard


C-Love:
With the 2023 Legislative changes, there is an update to how some GRH cases are budgeted. There is a new disregard for members in Supportive Housing facilities that have unearned income. This new disregard replaces other disregards in the budget and generally allows residents to keep more of their income.

To support this change, the case needs to be processed with a manual FIAT to change the disregard. This can only be done for residents who meet the following criteria:
- Have any unearned income
- Reside in a community supportive housing setting (MONY/VND2 Health Dept Licensing Code: 10)
- Eligible under SSI, AGED, ADULT, BLIND, or DISABLED basis only (MFIP and TRTX no elig).

This field about community supportive housing is not often viewed and is kind of 'hidden'. This means it would be easy to miss these cases and fail to manually FIAT as appropriate. Since the Eligibility Summary script is already reviewing vendor details and the budget, we should be able to programmatically identify cases that meet the criteria. 

This functionality would stop Eligibility Summary from operating on GRH cases that appear to meet the supportive housing disregard but it has not been applied. The case would need to be updated, reapproved and the script run again to support the advancement of awareness and quality checks. This would be user focused and not a background process.

C-Love:
Email sent to MA and ST of the GRH team to confirm this use would be supportive of the process.

C-Love:
Error handling is accurate and in place. Resolved.

C-Love:
Reopening due to procedural handling needed in the initial month of GRH as income is disregarded.

C-Love:
New issue resolved.

C-Love:
Past months bug found.

C-Love:
Discovered bug fixed.

PR #2124, opened 2024-12-12
#1952 NOTES - Elig Summ - Add GRH parameters for vendor info

C-Love:
Error because these parameters weren't defined to the class but are used outside of the class.


C-Love:
Merging as this is causing an error that the GRH team would like to test.

PR #2126, opened 2024-12-12
1898 NAV - FIND MMIS PMI IN MAXIS & FIND MAXIS CASE IN MMIS, update script functionality

MDR22:
Update both scripts to align with current scripting standards. Updated FIND MMIS PMI IN MAXIS to end at DSPL panel instead of navigating to CASE/NOTE. Updated the function navigate_to_MMIS_region to make it work with MMIS training region.

### DIFF
--- a/MASTER FUNCTIONS LIBRARY.vbs
+++ b/MASTER FUNCTIONS LIBRARY.vbs
@@ -11284,29 +11284,47 @@ function navigate_to_MMIS_region(group_security_selection)
 		If MAI_check <> "MAI" then EMWaitReady 1, 1
 	Loop until MAI_check = "MAI"
 
-	EMReadScreen mmis_check, 7, 15, 15
-	IF mmis_check = "RUNNING" THEN
-		EMWriteScreen "10", 2, 15
-		transmit
-	ELSE
+	EMReadScreen mmis_production_check, 7, 15, 15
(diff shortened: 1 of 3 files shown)


C-Love:
@MDR22 - Question about the training region support. Do both scripts support the training region? Or is it just MMIS to MAXIS? With that support, is the expectation that the session only has production OR Training, and not both?

MDR22:
> @MDR22 - Question about the training region support. Do both scripts support the training region? Or is it just MMIS to MAXIS? With that support, is the expectation that the session only has production OR Training, and not both?

Both scripts have training to training support and production to production support. The MMIS to MAXIS script has "simpler" handling for determining whether to navigate to training or production as it doesn't check all windows the way the MAXIS to MMIS function does. I could create a function for MMIS to MAXIS so that it operates in the same way. 

And yes, I kept the functionality where it will only work if production OR training is open, but not both. I couldn't think of way to determine whether the user is in training or production without navigating to SELF, which I thought might disrupt the user if they wanted to stay at a certain panel when navigating from MAXIS to MMIS.

C-Love:
Ok - I think I found the issue I was having. The redirects that the buttons are not set up to run locally. Once I ran it directly, it worked as expected for the training option once I ran it directly.

C-Love (approved):
Looks good to me!

MeganGeissler (approved):

Issue #1887, opened 2024-07-02
Reading STAT with a class structure is used in multiple scripts and needs alignment


C-Love:
Both HC Eval and Eligibility Summary use a class to read and save STAT information.

HC Eval has a separate file that is references.
Eligibility summary has much of the same code in the script file.

We need to align these different methods, it makes the most sense to have a separate file. 

IMPORTANT - the two classes are not the same at this point. We should ensure all of the information is read in this separate file.

C-Love:
It did not appear that I broke anything with this update.

PR #2128, opened 2024-12-13
#1887 Class stat-details all in one separate file

C-Love:
We already had a separate file for definition of the class stat-details for HC Eval to reference. Since we don't need to duplicate code all of the time, we should reference this file in Elig Summ. 

Incorporated parameters that are specific to Elig Summ into the MISC/class stat details file. 
Removed class stat-detail definition from Elig Summ
Added a call to the MISC/ Class Stat details file in Elig Summ to load the stat-details definition.

I want to pull this in at the beginning of the day because there is no way to test every scenario impacted by this change. If all goes well, workers should not notice this change.


MeganGeissler (approved):

courtrightd (approved):
хорошо. Approval granted.

Issue #2033, opened 2024-10-08
Add Usage Evaluation Parameter to CLoS


C-Love:
To better evaluate script usage, we need a way to group scripts by their 'use value'. To best track this detail, it should be recorded in the CLoS to allow access and updates. 

This is somewhat of a temporary solution as we resolve how to create a comprehensive script use report for performance management. The goal would be that this would eventually be available for SQL and PowerBI.

PR #2130, opened 2024-12-16
#2033 Assigned Usage Categories to CLOS

MeganGeissler:
Assigned Usage Categories to CLOS


C-Love (approved):
Looks good to me. Functioning well with the start I have to the reports.

Issue #2129, opened 2024-12-16
NOTES - EMERGENCY, update 12-month determination based on issuance date 


MDR22:
Update script to determine 12-month eligibility based on the issuance date, rather than the first of the month of the issuance.

PR #2131, opened 2024-12-16
#2129 NOTES - EMERGENCY, update 12-month determination based on issuance date 

MDR22:
Received feedback from user that emergency 12-month eligibility determination should be based on the issuance date rather than the first day of the month of the issuance (i.e. using 12/15/2023 instead of 12/01/2023). Made updates to script:

- Update changelog
- Updated 12-month eligibility determination to be based on issuance date rather than first day of month of issuance
- Fixed issue with screening_determination where it was incorrectly marking a case as not eligible when the case was applying for emergency assistance again, but AFTER the end of the 12-month eligibility period
- Fixed small typos/formatting


MeganGeissler (approved):

courtrightd (approved):
Looks great thank you!

PR #2132, opened 2024-12-17
#1952 NOTES - Elig Summ - automate email if script stops for GRH error

C-Love:
Added information to the error dialog and the ability to include notes about the case. 

The email will be automated to the script team. This email does not indicate that there is an error inherently, but just as a way to track these cases and this functionality.

The current information is that this FIAT and the subsequent error checking will be no longer necessary at some point in 2025, at which time all error handling can be removed. 

This will allow us more warning of future issues with this handling in using eligibility summary since the script is not designed to continue when this error is hit.


courtrightd (approved):
A-OK to merge.

Issue #2135, opened 2024-12-18
UC Verif Request - reverse order of DOB and SSN fields


C-Love:
Reported twice by the same worker that the dob is reporting out YYYY DD MM and the SSN is XXXX XX XXX - so each field is reversed but the numbers within it has the same order. 

Unclear how this is happening. Maybe we start with a better report out.

PR #2136, opened 2024-12-18
#2135 UTIL - UC Verif Req - report out on reversing

C-Love:
Maybe this can give us more info


courtrightd (approved):
I like it. Send it!

Issue #2148, opened 2024-12-30
ADMIN - DAIL CCD: increment dail_row in wrong place


IlseFerris:
Script needs update to have dail_row incremental variable in else statement instead of current placement.

PR #2150, opened 2024-12-30
#2148 ADMIN - DAIL CCD: Updated dail_row location

IlseFerris:
This needed to be in an else statement
Added changelog

tested and working as intended


courtrightd (approved):
Approved!

PR #2151, opened 2024-12-30
2149 fset exemption funclib

IlseFerris:
Tested and working well. I can give you a case or two (real cases) if you want to run it.


courtrightd (approved):
Approved!

Issue #2153, opened 2025-01-02
Update 10-day cutoff dates in FuncLib


MDR22:
See line 140 in FuncLib

PR #2154, opened 2025-01-02
#2153 Update 10-day cutoff dates for 2025

MDR22:
Update 10-day cutoff dates for 2025 to address error message that workers have been seeing. TE19.132.


courtrightd (approved):
Thanks for fixing so promptly! Approved!

Issue #2156, opened 2025-01-03
Add handling to Interview when case cannot be Edited


C-Love:
Read the case to determine if a CASE/NOTE or SPEC/MEMO can be created at the beginning of the run. This will alert the worker that the script cannot run through the end but all details will be saved. This is primarily to support cases that are out of county but need to interview in Hennepin County. 

The script would then indicate to the worker that they should run it again once the case is transferred, within the next week. All details will be saved and the script will end intentionally before any action point occurs.

C-Love:
Appears to have been used 3 times correctly. Appears to work!

PR #2159, opened 2025-01-09
#2156 interview inquiry handling

C-Love:
Read MAXIS CASE NOTE creation to determine if the case can be edited. 

If not, a warning is given and at the end the script will stop before any actions are taken. 

Save your work functionality will support future updates once access is granted.


courtrightd (approved):
Send it!

Issue #2160, opened 2025-01-09
Copy Case Data for Training has weird Excel Handling Errors


C-Love:
In trying to use the Copy Case Data for Training script, the Excel was working crazy. 

Reviewed the code and there is some odd and unclear handling. 

I have some quick fixes for this I am going to put in place.

C-Love:
All done with this

PR #2161, opened 2025-01-09
#2160 ADMIN - Copy Case Data for Training - excel handling

C-Love:
We should use functions to open files and save directly to the file. Workers will need to determine that they have write access to the file.


courtrightd (approved):
Approval Bestowed!

Issue #2139, opened 2024-12-23
DAIL scripts - ensure consistency across scripts for "last message" handling


MDR22:
Update DAIL scripts so that they have consistent handling for determining if the script has encountered the last DAIL message and whether it needs to advance to the next page of DAILs.

PR #2164, opened 2025-01-13
#2139 DAIL scripts - ensure consistency across scripts for last message handling in DAIL

MDR22:
Updated DAIL scripts so that they have consistent handling for determining if the script has encountered the last DAIL message and whether it needs to advance to the next page of DAILs.


C-Love (approved):
Looks good to me!

courtrightd (approved):
Approved!

Issue #2166, opened 2025-01-14
New Script - Interview Team Case Worklist Creation


C-Love:
Follow up on interviews completed by the Interview Team needs to be assigned to regular workers. 

This will access the XML files to pull case information into an Excel worklist.

C-Love:
The worklist creation works and no errors reported.

PR #2167, opened 2025-01-14
#2166 NEW SCRIPT - ADMIN - Interview Team Cases Worklist

C-Love:
New script to create an excel worklist of cases with an interview completed by and Interview Team worker using the Interview script.


courtrightd (approved):
Did not test all the way to list generation, but what I did worked well. Looks great, thank you!

Issue #99, opened 2020-07-09
Create DAIL scrubber for CSES & MEC2 messages regarding income


MiKaylaH22:
Jessica is going to add more information to the DAIL page with instructions on what to do with these messages. Jessica’s latest response.  her response "check to see if they are on UC and delete the dail" Dails- All dail messages are not listed in the manual. Most that are self-explanatory are not there. This particular message is just telling us that unemployment insurance has been reported. Like any other report of income HSR should check to see if it is being budgeted in the case and if not request verification or in this case internally request the information. 

Unemployment Page[https://dept.hennepin.us/hsphd/manuals/hsrm/Pages/Unemployment_Insurance.aspx] in the HSR manual has a list of liaison’s

Error occurred on 6/17/2020 at 5:45:33 PM 
Error type - DAIL - add support for this DAIL message. 

Information: MEC2 NOT SUPPORTED IN DAIL SCRUB 
Script has issues/concerns in the following areas: 
 - CASE/NOTE 
Closing message: You are not on a supported DAIL message. The script will now stop. 
The message reads: UNEMPLOYMENT INS REPORTED TO MEC² WITH A START DATE OF 
Please send an error report if you would you like this DAIL to be supported.

IlseFerris:
https://hennepin.sharepoint.com/teams/hs-es-manual/SitePages/MEC2.aspx - HSR Manual link to MEC messages

IlseFerris:
CS REPORTED: NEW EMPLOYER message are procedurally more in-line with MEC2 messages than new hires messages. Adding this support to this issue.

IlseFerris:
JA and I are putting together a series of questions about the new hire process and the associated interfaces (like these messages) for DHS. This procedure is essential for #514.

PR #2176, opened 2025-01-21
#99 Create DAIL scrubber for MEC2 messages

MDR22:
Added functionality to DAIL scrubber to determine if MEC2 message is actionable or non-actionable. For non-actionable messages, the script will notify the worker before deleting the message. The script has handling to navigate back to the DAIL message if needed to delete the message. If the MEC2 message is actionable, then the script will navigate to the corresponding STAT panel and provide a link to the HSR manual for further guidance on how to process the message.

### DIFF
--- a/COMPLETE LIST OF SCRIPTS.vbs
+++ b/COMPLETE LIST OF SCRIPTS.vbs
@@ -3060,6 +3060,22 @@ script_array(script_num).usage_eval				= ""
 script_num = script_num + 1							   'Increment by one
 ReDim Preserve script_array(script_num)	   'Resets the array to add one more element to it
 Set script_array(script_num) = new script_bowie	  'Set this array element to be a new script. Script details below...
+script_array(script_num).script_name		    = "MEC2 Message"
+' script_array(script_num).description		    = "Deletes non-actionable MEC2 messages and opens a dialog with policy links for actionable MEC2 messages."
+script_array(script_num).category               = "DAIL"
+script_array(script_num).workflows              = ""
+script_array(script_num).tags                   = array("")
+script_array(script_num).dlg_keys               = array("")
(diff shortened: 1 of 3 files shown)


dail/MEC2-message.vbs
```diff
@@ -0,0 +1,487 @@
...
+						End if
+
+						'Reset dail_row to 6
+						dail_row = 6
+					End If
+				Loop
+			End If
+		End If
+	End If
+End if
```
C-Love, line 439: I got a run to happen where the script never hit a script_end_procedure, so it ran back to the DAIL scrubber and told me the message was not supported.

It think this happened on a non-actionable MEC2 message where I stayed on DAIL so the IF statment on line 151 skipped all of the deletion logic.

C-Love (changes requested):
The actionable run worked well. I hit a bug with the non-actionable.

COMPLETE LIST OF SCRIPTS.vbs
```diff
@@ -3057,6 +3057,22 @@ script_array(script_num).used_for_elig			= False
 script_array(script_num).policy_references		= array("")						'SEE Line 58 for format'
 script_array(script_num).usage_eval				= ""
 
+script_num = script_num + 1							   'Increment by one
```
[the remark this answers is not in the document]
MDR22: Fixed - thanks!

dail/MEC2-message.vbs
MDR22: Great catch. I added so much handling for instances where someone navigates away from the DAIL that I didn't include handling for when they didn't. Should be fixed now. I also added handling for the script to end if it unable to find the matching message for some reason.

PR #2179, opened 2025-01-22
#2166 ADMIN - Interview Team Worklist - Allow current day interviews on the list

C-Love:


courtrightd (approved):
Thank you for this update, approved!

Issue #1920, opened 2024-07-26
MISC - FAA HEALTH CARE INFORMATION REPORT: Enhance to add secondary search criteria 


IlseFerris:
Request from FFA user, SS: 

> Due to residents having multiple PMI numbers, the PMI number listed in SSIS may not be the PMI number with active Medical Assistance.  When this occurs, the County misses out on revenue by being unable to bill for billable services.  We are looking for a tool to assist us with searching MMIS by name and date of birth, to find additional PMI numbers so we can update SSIS to reflect heath care status more accurately.  Currently we are manually performing the name and date of birth searches and recording the findings in Excel.  We have discovered this is a time intensive process which could potentially be improved with the use of Blue Zone Scripts.   
>
This tends to happen the most with minor children that have PMI's created without SSN's associated (which shows up as SSN's with all 0's). By creating an if...statement where if the SSN's isn't present to research the PMI under the last name, first name and DOB as listed in MMIS. This will patch more of the HC eligibility that will support this request. There are some limitations though (there may be more criteria once the work starts): 
- Last names need to be at least 3 characters, this will need to be added as a fail parameter.

IlseFerris:
Additional request from team was to add denial reason and code to the 1st elig span.

IlseFerris:
Met with SS, who is the primary user. She has asked for the following updates: 

- [ ] Output both the billed PMI and the RSUM PMI. 
- [ ] Output the name as shown in RSEL if more than one record found. 

Both of these updates are to help reconcile the records in SSIS and MMIS, and to distinguish which PMI was billed vs. additional records with new PMI's.

IlseFerris:
Met with SS, primary user. Her management has some data privacy concerns, and have asked for an indicator (Y/N) if the Billed PMI matches with a unique SSN (vs. 000-00-0000 SSN). This has been added, and tested. User is satisfied with the output and matching.

PR #2180, opened 2025-01-22
#1920 FAA HC Information Report: Post testing updates

IlseFerris:
Shortened 1st name read as to leave out the middle initial. Added RSEL PMI and name to output for record evaluation. Added if statement to determine which PMI to use.
Updated case status output text for accuracy.


courtrightd (approved):
Approved!

Issue #2178, opened 2025-01-22
NOTES - Interview - signature handling


C-Love:
Update the functionality in the Interview script for signature.
1. Remove dates for signature
2. Handle verbal signature better (use POLI references)
3. Update second signature handling

These updates will simplify the interaction with the signature details at the end of the interview script and support policy more comprehensively.

C-Love:
This is resolved and has made everyone's lives easier.

Issue #2158, opened 2025-01-07
Update Interview Verifications to include a Clear option


C-Love:
Allow worker to clear all verifications from the Interview script. 

Workers can already change verifications that are associated with a specific question, but this would remove all of them and allow the verifications selected in the verification dialog to be removed.

C-Love:
No errors reported, appears to be functioning.

Issue #2050, opened 2024-10-17
Interview - Expedited Determination Handling Enhancements 


C-Love:
As a part of the Interview and within the script, we need to be able to make an expedited determination when the interview is completed. Policy states that if we use the best information we have during the interview and the case appears expedited, we can process it as expedited with no issue. 

In order to have the best information available to the processing workers, the interviewers need to identify the information needed to make this determination. We need to embed the questions as supplemental questions into the regular interview process so the expedited determination is not a separate process of the interview. 

The display of expedited determination will need to be adjusted for our interview team to make it easier to explain as the part of the interview.

C-Love:
This appears to be working as expected. No errors reported at this time. Complete!

PR #2182, opened 2025-01-24
#2050, #2158 & #2178 NOTES - Interview Updates to enhance user experience

C-Love:
Updates to the interview script to better support forms, interview policy and handling. 

- Verifications can be entirely cleared. 
- Improve the 'Clear' button for all question based verifications.
- Updated the verbiage in the dialogs and outputs to align with the the allowance for different form selections.
- Expedited Determination output in the WIF needed better logic to handle for cases that do not require an expedited determination and for the output when used by the interview team.
- Removed signature date. 
- Updated handling for verbal signature documentation.
- Default second signature to Not Required for a single person household.

This update does not handle for two additional planned updates:
1. Add Program History information to Follow Up (maybe XML output). 
2. Update the Person Display information dialog.
I will work on these depending on other incoming work in the next week.

### DIFF
--- a/MASTER FUNCTIONS LIBRARY.vbs
+++ b/MASTER FUNCTIONS LIBRARY.vbs
@@ -1662,7 +1662,7 @@ function access_ADDR_panel(access_type, notes_on_address, resi_line_one, resi_li
         addr_eff_date = replace(addr_eff_date, " ", "/")						'formatting the information from the second half
         addr_future_date = trim(addr_future_date)
         addr_future_date = replace(addr_future_date, " ", "/")
-		addr_future_date = DateAdd("d", 0, addr_future_date)					'Ensuring date is in date format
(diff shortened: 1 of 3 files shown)


C-Love:
Added program history detail to the program selection dialogs. Additionally tweaked the program selection dialogs to be slightly different for the interview team workers vs standard workers.

MDR22 (approved):
The changes work well - looks good! I did notice an issue with the CAF ADDR screen in the dialog - looks like a 65th element issue. 

image

courtrightd (approved):
Tested good for me, looks good! I did notice the 65th element issue mark pointed out, but don't think we can do anything about it, and it isn't really inhibiting.

C-Love:
Yes, this is a known 65th element issue on the Member portion. We have an issue to address the member details in the coming months.

Issue #2195, opened 2025-02-03
Eligibility Summary - add error check for cases coded incorrectly for six-month budgeting workaround


C-Love:
Add handling to review cases for:

- Suspended MFIP cases.
- SNAP cases with monthly reporting 
- GA cases with monthly reporting

Script will not proceed under these conditions. Error messages will need to be sent to policy support team/person. 

Additionally, will add a line to MFIP cases that are listed as monthly reporting about policy change. 
UHFS and GA cases with earned income will also need a line added to the NOTE about the policy change for CASE/NOTE clarity.

C-Love:
Functionality is all in place and working

PR #2197, opened 2025-02-04
#2195 NOTES - Elig Summ - Added error checks for new 6-month reporting

C-Love:
Updates:
Error flag for suspended MFIP approvals
Error flag for monthly reporting on UHFS cases
Error flag for GA monthly reporting
Added WCOM for GA approvals with earned income


courtrightd (approved):
Looks good to me, once change-logged send it!

Issue #2192, opened 2025-01-31
ACTIONS - Earning Income Budgeting - Update to Support Six-Month Budgeting Workaround


C-Love:
The update to the JOBS panel has a workaround to support MFIP, GA, and UHFS cases. Details of this process can be found on SIR: https://www.dhssir.cty.dhs.state.mn.us/Shared%20Documents/Guide%20to%20Six-Month%20Budgeting%20Final%20Version.pdf

Essentially, the anticipated monthly average is entered as a single paycheck on the main JOBS panel in both retro and prosp sides to immitate budgeting sourced from the PIC even though there is no PIC to support these programs.

C-Love:
Updates appear to be working well.

C-Love:
The navigation of the PIC to allow for more than 5 checks was causing a problem with reading the total for the single payment functionality. Updated to scroll down ONLY if a 6th check is being entered. 

Basically the PIC was closing before we could read the calculated average and number format error was throwing.

C-Love:
No additional issues reported, script appears to work correctly.

PR #2199, opened 2025-02-04
#2192 ACTION - EIB - resolve logic errors in workaround selection

C-Love:
My logic got a little twisted up here but this should have a better selection.


courtrightd (approved):
Looks good to me! Proceed with merge!

MDR22 (approved):
Ran through a few times with no issues. Looks good!

Issue #2189, opened 2025-01-29
NOTES - HRF, update to reflect new policy guidance


MDR22:
Make updates to HRF script to reflect new DHS policy guidance:

- Add language to dialog to make it clear that this is the last month, OR add first dialog with this information before main dialog
-  Include a link to the new policy/guidance
-  Add language to CASE/NOTE to make it clear that there is a change
- Recreate the functionality to read JOBS, BUSI, and UNEA panels
- Error checking for JOBS panels 
- Plan to retire April 2025

PR #2200, opened 2025-02-05
#2189 NOTES - HRF, jobs and unea panel reading error handling

MDR22:
Updated error handling for JOBS and UNEA panels:
- Fixed error where JOBS error handling was updating UNEA error message instead of JOBS error message
- Fixed handling for both JOBS and UNEA panels where blank fields are replaced with "0.00" so that it can properly match across prospective, retrospective, and SNAP PIC fields
- Added improved handling for JOBS with income end dates. Script will check if the end date is from a previous month, if so then retrospective, prospective, and SNAP PIC all need to be set to 0.00 or blank. If the JOBS income end date is the first of the month and it matches the footer month, then the panel will be evaluated to make sure that retrospective, prospective, and SNAP PIC are all the same amount.

### DIFF
--- a/notes/hrf.vbs
+++ b/notes/hrf.vbs
@@ -1575,7 +1575,7 @@ function HRF_add_JOBS_to_variable(variable_name_for_JOBS)
 	' date_of_pic_calc = replace(date_of_pic_calc, " ", "/")
 	transmit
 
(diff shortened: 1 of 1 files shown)


courtrightd (approved):
Seems to be good to go from here. Approved!

Issue #2187, opened 2025-01-28
NOTES - INTERVIEW, resolve ADDR panel issues


MDR22:
Users reported issues with NOTES - INTERVIEW when it attempts to update ADDR panel related to the Address Eff Date and Address Verification fields. Will update the script to ensure it does not write a blank verification to ADDR panel and also update access_addr_panel function to ensure it is comparing dates.

PR #2201, opened 2025-02-05
#2187 NOTES - Interview - safety measure for addr_eff_date

C-Love:
Error reported when script is run from save your work.


MDR22 (approved):
Makes sense. Tested and didn't hit any issues.

courtrightd (approved):
Approved

Issue #2208, opened 2025-02-12
#671 DAIL - UNCLEAR INFO, ensure CM used for ELIG/FS


MDR22:
Resolve error with DAIL - UNCLEAR INFO runs where the script uses the first DAIL message month for a case to determine if the case is a SNAP 6-month reporter from ELIG/FS instead of using current month.

Issue #671, opened 2021-12-09
SNAP Six Month Reporting Case Action/DAIL Supports - Unclear Info


IlseFerris:
During a recent FNS audit, auditors identified that as a state we are not processing HIRE messages for SNAP only cases that are six month reporters correctly. During a period of discovery, we also believe that this would apply to CSES messages as well. 

Deliverable: Create a BULK script to be run on the 4th (or closest working day after the 4th) of each month that would pull a list of HIRE and CSES messages, and identify which cases are SNAP only and are six month reporters. Those cases would then have the DAIL messages processes and/or deleted and captured, based on the type of message. 

Several questions still need to answered before a final plan can be created. Jennifer Frey is sponsoring this project and is finding out the answers to the following questions: 
- [ ] Federal new hire matches: what options would we select/coding for the interface? What should a case note look like? Also guidance for closed cases with matches sitting out in MAXIS?
- [ ] If a letter is sent from the agency informing residents that a HIRE match was received and reminding them of their reporting responsibilities, is this considered a negative verification?
- [ ] FNS language in writing regarding interface matches: can we disregard matches? CSES matches acting the same way?
- [ ] FNS CAP and steps DHS is taking to mitigate findings.

IlseFerris:
AMT CHILD SUPP MOD/ORD = $732.00 FOR CHILD REF NBR: 05 - Look at this support.

PR #2209, opened 2025-02-12
#671 #2208 DAIL - UNCLEAR INFO, Ensure ELIG/FS viewed for CM, add changelog

MDR22:
Ensure ELIG/FS is viewed for current month rather than defaulting to the footer month for the first DAIL message encountered for a case.


courtrightd (approved):
Looks good to me!

Issue #2221, opened 2025-02-28
MFIP Orientation output to XML has a bug in the index


C-Love:
Script error when the MFIP orientation functionality is triggered in the XML. Index not correctly aligned. 

Additionally added the windows user id so we can capture user name if the staff view glitches again.

C-Love:
Appears to be resolved.

PR #2222, opened 2025-02-28
#2221 NOTES - Interview - align the index in the xml output

C-Love:
Confirmed this should be enough with a live test from the worker AND at test script.


courtrightd (approved):
Looks good to me

Issue #2236, opened 2025-03-11
NOTES - INTERVIEW FORMS CLASSES, adjustable text boxes


MDR22:
Worker reported issue with PDF generated as part of interview script that cut off worker entered comments regarding JOBS panel. Add handling to adjust cell height based on length of worker comments

PR #2237, opened 2025-03-11
#2236 MISC - INTERVIEW FORMS CLASSES, Adjust cell height based on length of worker comments

MDR22:
Worker reported issue with PDF generated as part of interview script that cut off worker entered comments regarding JOBS panel. Added handling to adjust cell height based on length of worker comments


courtrightd (approved):
Looks good thank you!

Issue #2225, opened 2025-03-03
Case Sampling structure and information automation


C-Love:
A case sampling process has been created that requires a number of manual steps to manage the process.

1. PND2 from previous day and current day are compared
2. Case selection criteria: pending ended (cash and snap only, on list yesterday, not on list today), cases pended for at least 20 days, apparent denial with I code excluded)
3. A list separated by population is generated.
4. Workers then create a case sampling document for each case from the list using a template.
5. After all reviews are done, details from the case sampling document are copied into a compilation document.

These steps can all be completed using script automation.

C-Love:
This is working and tested.

PR #2245, opened 2025-03-19
#2225 ADMIN - Case Sampling - compilation file handling tweaks

C-Love:


C-Love:
Pulling in for the morning run. Single user script should be well supported for testing.

PR #2224, opened 2025-03-03
#2195 & #2223 NOTES - Elig Summ 

C-Love:
Updates to email notifications in Elig Summ:
- Six Month budget errors to go to supervisor
- Created function to read supervisor email from REPT/USER and the employee view
- Changed the QI member who receives 30/09 errors

PR #2246, opened 2025-03-20
#2224 NOTES - CAF, Improve handling for reading/adding Verifs CASE/NOTEs

MDR22:
Add handling to ensure that script creates a new Verifications Requested CASE/NOTE when new verifications are added to the Verifs Needed field in the dialog. If no new verifications are added, then the script will not create a new CASE/NOTE.


courtrightd (approved):
Looks like a good solution to me! Thank you!

Issue #2248, opened 2025-03-24
BULK - Rept EOMC list Bug Reported


C-Love:
The navigation functionality in BULK - REPT EOMC threw an error when the loop to read each worker exited without blanking out MAXIS_case_number. 

Moving the navigation outside of the loop may work but it could introduce issues with moving between workers. Adding supports to ensure MAXIS_case_number does not 'catch' the navigation.

C-Love:
This is fixed.

PR #2249, opened 2025-03-24
#2248 BULK - REPT EOMC list Bug in Nav getting stuck on a case

C-Love:
This is in response to email from IF. It appears there is a way for the loop to exit without clearing the case number variable. There have been past issues if we leave the REPT on accident during the loop so I didn't want to pull the navigation out of the loop. 

I tested this run but the same error wasn't occurring this month so I can't confirm for sure that this is a resolution.


MDR22 (approved):
Makes sense and looks good to me!

courtrightd (approved):
okie dokie

Issue #2254, opened 2025-03-25
Bug on Interview for Incomplete functionality


C-Love:
The dialogs are looping weird with a lot of extra questions when the worker presses the interview 'Incomplete' button. 

Worker reported this error when they canceled instead of using the incomplete functionality.

C-Love:
This is working! Yay better dialog navigation.

PR #2255, opened 2025-03-25
#2254 NOTES - Interview - dim the variable to leave the loop

C-Love:
The main dialog kept appearing when the incomplete option was pressed. Plus there were two confirmations before leaving the loop. 

Worker reported this prevented them form using the incomplete functionality.


courtrightd (approved):
Ope! That should do it.

Issue #2258, opened 2025-03-27
Interview BUG when updating PROG if verbal program requested


C-Love:
Reported by LD Team that the panel froze the script run when trying to add the verbally requested program interview date. 

Typically the script prevents this with a dialog to select which programs to update. With the LD Team we need more handling to adjust for process steps. Additionally, we don't want the script to stall.

C-Love:
Resolved - no issues reported.

Issue #2235, opened 2025-03-11
Add Early interview tracking for LD workers


C-Love:
In the case that the interview script errors after creating the CASE/NOTE - we may miss tracking cases that were interviewed. 

While we have reminded leadership of the importance of reporting script errors, these can be missed on accident. If we add a tracking method at the beginning of the script run, we can be sure to capture all cases. 

1. Create a new XML for interview start. 
2. When the interview is finished and the XML for the end of interview is created, we can delete the start file.
3. We can also add support to check for an already existing XML and delete it so we don't have duplicates
4. We also need to delete the start file if the script is ended as incomplete. 

This should ensure better worklist creation and case tracking.

PR #2259, opened 2025-03-27
#2235 & #2258 NOTES - Interview updates

C-Love:
Move LD Early tracking file to after main dialog.

Update PROG functionality to remove interview dates if program does not have an APPL date.


courtrightd (approved):
Looks good from here!

Issue #2243, opened 2025-03-17
ACTIONS - EIB, add conversation field to dialog that applies to both SNAP and MFIP


MDR22:
Worker requested addition to ACTIONS - EIB to include a conversation/clarification field to dialog that would apply to both SNAP and MFIP

PR #2262, opened 2025-03-28
#2243 ACTIONS - EIB, add conversation field to dialog that applies to both SNAP and MFIP

MDR22:
Moved conversation/clarification field to bottom of dialog regardless of programs. Also adjusted the sizing of the GRH groupbox to better fit contents.


courtrightd (approved):
Tested good for me! Please add a changelog before merge.

Issue #2267, opened 2025-04-02
DEU - PARIS MATCH CLEARED, show full resolution status in CASE/NOTE header


MDR22:
Request from worker to include the full resolution status in the CASE/NOTE header rather than the abbreviation for clearer communication purposes to other teams. Will update the script to show the full resolution status in the CASE/NOTE header.

PR #2268, opened 2025-04-02
#2267 DEU - PARIS MATCH CLEARED, show full resolution status in CASE/NOTE header

MDR22:
Request from worker to include the full resolution status in the CASE/NOTE header rather than the abbreviation for clearer communication purposes to other teams. Updated the script to show the full resolution status in the CASE/NOTE header.


courtrightd (approved):
Simple enough. Thanks!

Issue #2270, opened 2025-04-03
Interview Form - Question missing in dialog


C-Love:
Reported from a couple workers that Q12 in the CAF and Q11 in MNBenefits is not displayed in the dialog. 

It appears this is only happening when the script information is restored after save your work is used.

C-Love:
This is fixed! Group effort.

PR #2271, opened 2025-04-03
#2270 Interview Forms Classes - add answer_is_array parameter where missing

C-Love:
I tested this on the training case that was reported in error. 

I also checked the other form questions to see if it was missing anywhere else and didn't see anything.


MDR22 (approved):
Tested a few times and didn't encounter any missing questions. Looks good!

Issue #2274, opened 2025-04-09
Interview Update to allow for no MEMO permission


C-Love:
Interview script is failing if worker does not have the ability to MEMO. Since the MEMO is not strictly required, allow for script to continue if there is no update access for MEMO. Handling already in place for PROG/REVW updates to be ignored in this situation.

C-Love:
Confirmed this handles for the access of the worker who was having difficulty. No other issues reported.

PR #2275, opened 2025-04-09
#2274 NOTES - Interview - skip MEMO if not opened

C-Love:
Since I cannot turn off my MEMO access, I cannot test that option. However, this worker is on the interview line and should be able to check it pretty quickly once in place. As far as we are aware this is only impacting one person.

I did test that the MEMO still sends since I do have access.


courtrightd (approved):
Looks good thank you!

Issue #2279, opened 2025-04-15
NOTES - Interview - handling needed for exp determination amounts


C-Love:
The interview script is throwing an error for the LD team if the information entered into the fields for income, assets or housing on the Expedited page are not numbers. The script needs to complete mathematical operations on these values. 

Add dialog error handling to ensure these amounts are numbers.

C-Love:
Resolved!

PR #2280, opened 2025-04-15
#2279 NOTES - Interview - add dialog err handling for LD expedited amounts to be numbers

C-Love:


courtrightd (approved):
Perfect thank you!

Issue #2278, opened 2025-04-15
ACTIONS - EIB, fix dialog sizing error


MDR22:
Worker reported issue with JOBS dialog cutting off, likely due to addition of clarification fields and dialog not properly resizing to account for this. 

![Image](https://github.com/user-attachments/assets/4765d91e-d1a7-439a-8056-42ecb101e341)

PR #2282, opened 2025-04-16
#2278 ACTIONS - EIB, Update dialog sizing and coordinates

MDR22:
Changed dialog sizing and coordinates to ensure dialogs do not cut off fields.


courtrightd (approved):
Looks good thank you!

PR #2288, opened 2025-04-17
#2279 NOTES - Interview - BUG in getting Exp Det accurate

C-Love:
There's a bug in the expedited information because I used the wrong conditional to make sure the expedited assessment happened at the right time.


courtrightd (approved):
makes sense, merge at will!

PR #2289, opened 2025-04-18
#2278 ACTIONS - EIB, Fix dialog sizing, add changelog

MDR22:
Attempt 2 to make the dialog sizing work properly. It should work and resize now regardless of the number of checks and programs selected, but definitely worth double-checking.


courtrightd (approved):
Tested ok for me!

Issue #2233, opened 2025-03-06
Interview Team Worklist Update


C-Love:
The worklist cases need to be added to a shared assignment excel file.
A assignment rotation system has been created and should be able to be automated.

C-Love:
This has been tested by both users and is working.

PR #2290, opened 2025-04-18
#2233 ADMIN - Interview Cases Worklist - add excel formula if missing

C-Love:
If this doesn't copy correctly in the Manager Log excel - then the processor assignment is blank. 

Basically a backup for excel doing weird things.

Tested this morning.


courtrightd (approved):
looks good from here!

Issue #2299, opened 2025-05-02
NOTES - INTERVIEW, adjust button locations


MDR22:
Feedback from worker: See screenshot to follow. On the dialog for updating verifications, the clear verifications button is at the bottom right. On most scripts, the button at bottom right is typically for confirming the current dialog or the script, or in some cases a cancel button which is not final and will not clear much data that has been entered. For me, this has sometimes contibuted to accidentally clearing verifs I didn't mean to clear, because the UI is not consistent with the UI norms of many other scripts. Suggestion is to move the clear verifs button.

![Image](https://github.com/user-attachments/assets/3393eda8-4c70-409a-9f54-93c4c8e46ad0)

PR #2300, opened 2025-05-02
#2299 NOTES - INTERVIEW, Update location of verif buttons to align with other dialogs

MDR22:
Updated locations of Verifications Clear button and the return to dialog buttons to align with other dialogs.


courtrightd (approved):
Looks good thank you!

Issue #2302, opened 2025-05-05
NOTES - HRF, remove all programs save for MSA


MDR22:
Remove all programs except for MSA on initial dialog and handling within the script.

PR #2303, opened 2025-05-06
#2302 NOTES - HRF, remove all programs save for msa

MDR22:
Updated NOTES - HRF to only include MSA checkbox on initial dialog and then removed functionality for all other programs except MSA.


courtrightd (approved):
Looks good, tested goo. Thank you!

Issue #2307, opened 2025-05-12
NOTES - INTERVIEW, UNEA income details mixed up in CASE/NOTE


MDR22:
According to worker: "I've noticed that when I select "Yes" for SSI while completing the Interview BlueZone script, it appears as "UI" in the final case/note (it appears the SSI fields and UI fields are swapped)." I confirmed that the script does mix up the values entered in the dialog.

PR #2308, opened 2025-05-12
#2307 NOTES - INTERVIEW, UNEA income details mixed up in CASE/NOTE 

MDR22:
Resolved issue with script writing UNEA income details from dialog in CASE/NOTE but mixing up the values.


courtrightd (approved):
Looks good thank you!

Issue #2314, opened 2025-05-19
Case Sampling - Create BiWeekly Report functionality


C-Love:
Add the ability for the Case Sampling script to create a time period report out that generates an Excel of case that need to be repaired for assignment with very limited details. 

Additional an Excel of all the cases from the time period will be created with a reduction in the number of columns available for viewing. 

This will run as a separate option and is not incorporated in the daily run options. This work is provided primarily to JF.

PR #2315, opened 2025-05-21
#2314 ADMIN - Case Sampling Biweekly Report Creation

C-Love:
Added an option to create a compiled report on an interval for fixing cases and evaluating reviews.


courtrightd (approved):
okie dokie!

Issue #2295, opened 2025-04-25
BUG on NOTES - Interview - HUF Q7 notes missing


C-Love:
Reported that the HUF Q7 wasn't saving any information from the interview notes.

Actually the request was for an additional field for LIHEAP but in responding about the notes, I realized the notes didn't correctly copy into the MAXIS output.

First issue is to resolve the notes issue. 

Additional discussion about expanding Q7 to include LIHEAP specific question.

C-Love:
This did NOT break the interview script. So should be good.

PR #2316, opened 2025-05-21
#2295 MISC - Interview Forms - fix logic for CASE/NOTE entry HUF Q7

C-Love:
Updated the logic in the forms class to:
- Correctly identify if the questions is 'entirely blank'
- Ensure entry of details for shel-hest question types

This should resolve missing detail from the error report.


courtrightd (approved):
Looks good from here.

PR #2318, opened 2025-05-22
#2283 Rent-Help hennepin memo

courtrightd:


C-Love (approved):
Simple and clean. Functionality looks good to me.

For awareness, on EGA pending only cases the interview notice still goes out. I don't think this is an issue but good to know as a part of the functionality

MDR22 (approved):
Looks good to me

Issue #2321, opened 2025-05-23
ELIG Summ - HC EMA approvals have wrong test


C-Love:
The test in ELIG/HC for EMA approvals has 'Emergency Exists' instead of 'Elig for Other Programs'.

Update to ELIG Summ to support correct documentation of these approvals.

C-Love:
Updates make - seems to be working!

Issue #2319, opened 2025-05-22
ELIG SUMM - HC Tests Changed


C-Love:
ELIG/HC has added a line on the Cooperation pop-up for the cooperation tests for AVSA. 

This has moved the rest of the cooperation tests down on the screen. 

This is causing the wrong tests to be indicated in ELIG/SUMM.

C-Love:
During discovery on this issue I am noticing that the MSP functionality does not open pop-ups on the Person Test panel. To achieve this would need to copy the do loop code block on 18448 to line 20011 and then tweak for the specific pop-ups on MSP programs. 

This is not really a bug so I haven't included it automatically. For discussion or future work?

C-Love:
Fixes in place!

Issue #2181, opened 2025-01-23
Retire GRH NON-HRF Post-Pay - Eligibility Summary to support Functionality


C-Love:
GRH can be paid in a post-pay method where approvals are completed after the month of eligibility. 

Currently GRH Post-Pay is being supported by a separate script. Since this process is only for the approval, Eligibility Summary should be able to handle this documentation. 

GRSM has a code 'Pre/Post Pay Code' to indicate that an approval is a post pay situation. 

Information in the GRH NON-HRF Post_Pay script to review.

May need to include: 
- FACI and ADDR comparison and details of address
- IAA status and information

Already in Eligibility Summary: 
- Income (earned and unearned)
- Postpay Results

No longer needed:
- Other notes
- Changes
- Verifs
- Action Taken

Need to confirm with GRH team and 1800 for functionality supports. Eligibility Summary has more comprehensive supports.

C-Love:
This should be functioning well. If there is other feedback on this update we can address it at that time.

PR #2322, opened 2025-05-23
#1885 & #2181 & #2319 & #2321 - ELIG Summ Bug fixes from user reports

C-Love:
Handling updates for the following:
- Update GRH Header to include payment processing detail (Pre-pay, Post-pay)
- Fix Cooperation tests since AVSA cooperation has been added.
- Fix accessing RETRO months
- Updated EMA person tests to include Emergency Exists, which was missing.

All HC tests were reported by the same worker and responses have been sent. 
GRH update information sent to the GRH team. 

Tests completed on the cases provided an a small sampling of other cases.


C-Love:
I looked at that message box DOZENS of times today. Never saw it.

Issue #2330, opened 2025-06-03
Application Inquiry - update email


courtrightd:
Utilities - Application Inquiry need to be updated to have the requests handled by case assignment. 

Case Assignment will be responsible for processing the application inquiry requests.  

Can you please update the script to have these requests sent to the following email and any verbiage needed to remove QI and add Case Assignment? 

 HSPH.ES.CA.MNB

PR #2331, opened 2025-06-03
#2330 UTILITIES - APPLICATION INQUIRY, Update email and script info to CA from QI, update changelog

MDR22:
Updated email and language from QI to case assignment.

I will wait to merge until Monday.


courtrightd (approved):
Thanks you looks good!

Issue #2333, opened 2025-06-09
Elig Summ - Workers with HC only access need special handling


C-Love:
Additional workers have been added to the HC team that do not have access to all areas of MAXIS, particularly MONY. These workers do not have access to cash or snap and Elig Summ should be updated so these workers only access HC results. There are 27 workers to include in this update, details received from BT.

This is already in place for two workers and just needs to be expanded.

C-Love:
This is resolved, no issues.

C-Love:
UGH - I put in MX numbers instead of windows numbers in. Need to update.

PR #2335, opened 2025-06-10
#2333 NOTES - Elig Summ - Update the workers with HC approvals only

C-Love:
Issue reported by KW and names confirmed with BT. 

Specific testing for these workers will be challenging and will need to mainly occur once merged since it is such a specific set of workers that should see the change.


courtrightd (approved):
send it

Issue #2336, opened 2025-06-10
ADMIN - Basket Review - Expand to ACTV and REVW


C-Love:
To identify some basket issues, I expanded the Basket Review script. 

Added:
- Works on all caseloads in the county.
- Only adds caseloads to the Excel if there are cases in the caseload.
- Adds counts for REPT/ACTV
- Adds counts for REPT/REVS

Work is completed and this script has limited use.

PR #2337, opened 2025-06-10
#2336 ADMIN - Basket Review - add ACTV and REVS

C-Love:
Updates to the script that can review case counts by X-number.

Mostly used internally.


courtrightd (approved):
looks good

Issue #2272, opened 2025-04-03
Interview - update CAF questions to align with new version of DHS-5223


courtrightd:
New CAF version released dated 03/25. 

- Numerous questions have updated phrasing to improve resident understanding.
- Question 8 is a totally different question.
- Q9 JOBS is reformatted to 9 and 9a
- All questions 8 - 23 are re-numbered, or new questions.

Will need to update the interview-forms-classes file to align with new CAF. Interview dialogs may need updates to display correctly with new question formats.

Pull request should be timed to coordinate with new form entering ECF.

C-Love:
Issue reported in save your work with a Boolean not being set correctly

C-Love:
This is now working as expected.

PR #2344, opened 2025-06-23
#2272 MISC - Interview Forms Classes - CAF Q14 reword

C-Love:
Reported uncertainty around CAF Q14 - new question has a list of options that do not fit well.
image

Updated question to just reference earned income.


courtrightd (approved):
Looks like a good change

Issue #2116, opened 2024-12-09
HC Eval - Note header update for Ex Parte mixed households


courtrightd:
Simple update to change the case note header to identify mixed households for ex parte.

PR #2350, opened 2025-06-27
#2116 health care eval ex parte mixed household header change, secu bug fix

courtrightd:


C-Love (approved):
Looks good! I double checked to make sure that CASE/NOTE isn't read in another script and it does not appear to be so this shouldn't cause any unintended issues.

PR #2353, opened 2025-07-01
2340 func lib coordinate changes

C-Love:
Update to the Read UNEA function so the scripts using it do not break. 
Does not add new functionality to read the new CASH PIC at this time.

Also fixed the indenting because my brain hated it so much.


courtrightd (approved):
Send it!

Issue #2340, opened 2025-06-20
MAXIS updates for Cash Six-Month Reporting


C-Love:
Since monthly reporting no longer applies to MFIP and GA, a number of changes are being made to MAXIS to support processing. This change will go into effect 7/1/25 and will impact MAXIS retroactively to footer month 03/25 going forward.

Panels updated:
- STAT/JOBS 
- STAT/UNEA
- STAT/REVW 
- ELIG/MFIP
- ELIG/GA (Additional review needed for ELIG/GA - unclear if changes exist here)

A couple of other notes on this update:
- RCA is also changed but this should not impact our processing.
- There are possible changes to GRH/HS and UHFS but these programs already support six-month budgeting so there should be minimal system changes. 

The scripts that need updating are:

- [ ] Eligibility Summary
       - Coordinates changed in ELIG/MF
       - Coordinates changed in STAT/JOBS and STAT/UNEA
       - Add logic for 'Prior Low' evaluation and documentation
       - Update logic for counted income with new PIC for MFIP and GA
       - Add logic for SR completion in MFIP and GA in the review completion at the end
- [ ] CAF 
       - Coordinate and logic changes for JOBS income
       - Coordinate and logic changes for UNEA income
- [ ] Earned Income Budgeting
       - Coordinate changes for STAT/JOBS
       - Add logic for Cash budgeting similar to SNAP
- [ ] Changes reported/Job Change Reported - review functionality.
- [ ] HC Eval and other HC script the read COLA Disregard
- [ ] Review code to ensure no required updates are missed.

The provided documentation is on SIR:
[ntroduction to Six-Month Budgeting for Cash Programs/UHFS](https://www.dhssir.cty.dhs.state.mn.us/MAXIS/trntl/MAXIS%20Resources/Additional%20Resources/Introduction%20to%20Six-Month%20Budgeting%20for%20Cash%20Programs%20UHFS%20-%20SIR.pptx?d=wcb828f3b3bef48d7bc2590cd87108c08)
[...]

PR #2354, opened 2025-07-01
#2340 NOTES - Exp Det - UNEA SNAP PIC selection moved

C-Love:
The UNEA panel was updated and in months 03/25 and onward has the SNAP PIC selection field moved.


courtrightd (approved):
Looks good, thank you!

Issue #2360, opened 2025-07-07
Title of SNAP REVW Pop-Up Changed


C-Love:
Effective footer month 03/25, the title of the pop-up has changed from 'Food Support Reports' to 'SNAP REPORTS'. 
Additionally, it may have moved for all months to be at the coordinates 5, 33 from 5, 30 - but this is hard to confirm. 

Many of the scripts that access this functionality are built to run around the current month and so conditional handling is not needed. 

The scripts addressed in this pull are:
- ADMIN - DAIL UNCLEAR INFO
- ADMIN - MONT REPORT
- ADMIN - REVIEW REPORT

The update needed in the Training Case Creator will be resolved in the issue #2358 and not handled here as conditional logic will be needed and the Training case creator updates are less urgent.

C-Love:
Additionally, the pop-up for the Cash/GRH REVW has changed. From 'GRH Reports' to 'CASH Reports'. This only impacts MONT Report and Review Report.

PR #2361, opened 2025-07-07
#2360 Pop-Ups on REVW Title Changes

C-Love:
Since cash six-month reporting is now working in MAIXS, there have been some more cosmetic changes that have happened to the REVW pop-ups. 

These changes are footer month specific but since the affected scripts work based on the current footer month - no conditional logic is needed. 

Currently operating well in DAIL Unclear Info - the other scripts will be tested later this month when new reports are created.


courtrightd (approved):
Looks great thank you!

PR #2362, opened 2025-07-07
#2333 - NOTES - ELIG Summ - USE THE RIGHT WORKER NUMBERS

C-Love:
The user id should be the windows user ID and not the MX one.


courtrightd (approved):
Ope! Fix looks good. Approved.

Issue #2339, opened 2025-06-12
BZ Database migration


courtrightd:
BZ Statistics database is moving to a new server. Calls to the database will need to be updated to the new path in multiple scripts.
Planned approach is to create a variable for the path in global variables so the change can be made in a single location.

- [x] Create global variable for BZ database path
- [x] identify scripts that utilize db connections and update to utilize the new variable
- [x] change global variable on migration date (7/20/25)

PR #2380, opened 2025-07-25
#2339 Add default for DB String for direct play of scripts

C-Love:
When we run scripts directly (instead of through the menus) the global variables are not read. 
This causes issues with the db being defined only in global variables. 
Added an if statement to scripts that require the db information to read the global variables on the c drive to fill this information. 

NOTE - I did not include EX PARTE scripts since I am not sure if you have updates to them. I can add them here or you can do it in your own branches/updated files.


courtrightd (approved):
Added a commit with ex parte report. Approved.

Issue #2395, opened 2025-08-19
BUG - Interview confirm interview done


C-Love:
If the interview is attempted to be completed from the main dialog and the script loops back after the worker confirms once. 

This is a minor error but an inconvenience. 

From Email on 7/17/25 from JP.

C-Love:
This is resolved.

PR #2398, opened 2025-08-20
#2395 NOTES - Interview - reset leave_loop on main dialog

C-Love:


courtrightd (approved):
That looks more better. Approved.

Issue #2393, opened 2025-08-19
Update Health Care PM


C-Love:
BT is moving to operations and FRC will be taking over the Program Manager role in Health Care. 

Updates needed to scripts that support Health Care administrative actions. 

The training alert for Eligibility Summary will go to JA.

C-Love:
Management updates are done.

PR #2399, opened 2025-08-20
#2393 Add FRC as HC PM

C-Love:


courtrightd (approved):
Looks great thank you!

Issue #2358, opened 2025-07-03
Fix Training Case Creator and Copy Case Data for Training with CASH SR Updates


C-Love:
There have been updates to MAXIS panels and the script to interact with these should be reviewed and updated to operate correctly. 

This is limited use but helpful to trainers and script writers. 

Related to work on #2340.

C-Love:
Fixed!

PR #2404, opened 2025-08-22
#2358 ADMIN - Training Case Creator - Resolve 2025 MX updates

C-Love:
Updates to Training Case Creator to Align with new MX updates. 
Tested on about 10 scenarios including MF and GA cases. 

It appears there is no need to update Copy Case Data for Training as it does not open any pop-ups.


courtrightd (approved):
Approved!

PR #2413, opened 2025-09-16
LTSS_statistics

IlseFerris:
Added stats collection to LTSS script set. 

Applied database and collection information to Global variables. 

Updated changelog, and added script_end_procedures to replace StopScripts


courtrightd (approved):
I bestow my approval upon thee!

PR #2416, opened 2025-09-22
#2393 - ADMIN - Ex Parte Report - Update ADMIN allowed users

C-Love:
Updates to the user ids to allow correct access to Ex Parte admin updates.


courtrightd (approved):
Approved thank you!

PR #2454, opened 2025-10-15
2452-shutdown-case-tracker

courtrightd:
Case tracking script that will note that SNAP or MFIP cannot be approved due to federal shutdown. we will utilize the end message in the stats database to "track" these cases and pull a report for follow-up in the future when we can do approvals.
Also includes a redirect file for new button on powerpads.


courtrightd (commented):
Self, I approve thee.

Issue #2431, opened 2025-09-30
Eligibility Summary - Update list of HC only workers


C-Love:
There are a list of workers in Eligibility Summary that can ONLY use the Health Care program functionality. This is to support the workers MAXIS access since some areas of other programs are not available to these workers. 

It appears there is an update to some of these workers. RZ has requested a one worker be removed from this list (PC). 

We will also need to check with the other HC supervisors to confirm who should be on this list.

PR #2461, opened 2025-10-22
#2431 & #2462 NOTES - Eligibility Summary - HC Only Workers and ELIG GA updates

C-Love:
Discussed with HC team to remove workers that now have access to SNAP/Cash in MX or workers who no longer work in Henn.


C-Love:
Added fix for another error report about bug on GA ELIG screen - this changed effective footer month 03/25.

Issue #2475, opened 2025-10-30
Bug in DAIL - COLA Revw and App when calling Elig Summ


C-Love:
HSR reported issue when using the COLA DAIL Scrubber script to call Eligibility Summary. It appears a name of a variable is redefined as a different type causing a type mismatch. 

Upon review it appears the variable 'name_array' is used in both scripts. 

Since the variable is used more often in Eligibility Summary and the script is larger, it is easier to update the name in the DAIL scrubber script.

C-Love:
This is resolved.

PR #2476, opened 2025-10-30
#2475 DAIL - COLA Review and Approve - Bug in running Elig Summ from button

C-Love:
Change the name of the variable 'NAME_ARRAY' so there is no conflict with Eligibility Summary.


courtrightd (approved):
Approved!

Issue #2472, opened 2025-10-29
ELIGIBILITY SUMMARY - MFIP Deemer Pop Up Header Change


C-Love:
Error reported in Eligibility Summary MFIP navigation from HSR AGK. 

It appears that the Deemer popup on MFB1 has another possible header that needs to be checked for.

PR #2481, opened 2025-10-31
#2472 NOTES - Elig Summ - multiple deemers header

C-Love:
Maybe there could be multiple deemers on a case that header needs to be looked for newly before the next loop. Missed this in the last update but we got a case reported to test on today.


courtrightd (approved):
approved

Issue #2492, opened 2025-11-04
DAIL - CSES SCRUBBER, temporarily deactivate script


MDR22:
Due to policy changes, deactivate the script temporarily by adding a script_end_procedure at the start of the script alerting users to the fact that the script is no longer available.

PR #2493, opened 2025-11-04
#2492 DAIL - CSES SCRUBBER, temporarily deactivate script

MDR22:
Added script_end_procedure to start of script alerting users to the fact that the script would be unavailable until further evaluation is completed related to the new budgeting rules.


courtrightd (approved):
Perfect thanks!

Issue #2414, opened 2025-09-18
NOTES - CSR, add verbal signature feature


MDR22:
Received feedback from user requesting addition of verbal signature feature - see request below:

Good morning, my suggestion is to add the verbal signature functionality to the CSR script, since it does not require an interview. The verbal signature functionality on the interview script is very good I think but since CSRs don’t have an interview, there’s not an automated way to document CSR verbal signatures. They have some specific requirements HSRs could miss that the interview script really helps with.

PR #2505, opened 2025-11-10
#2414 NOTES - CSR, add signature field and update script functionality

MDR22:
Made a number of updates to the script:
- Added a signature field to second CSR dialog (will only display verbal signature option if SNAP selected)
- Removed 
- Reformatted the dialogs and the script navigational buttons
- Removed the prev/next member and prev/next panel buttons as seemed to cause issues (i.e. inadvertently creating new panels)
- Utilized functions to generate dialogs
- Removed following checkboxes:
  1. "Check here to case note grant info from ELIG/FS.", grab_FS_info_checkbox
  2. "Check here if CSR and cash supplement were used as a HRF.", HRF_checkbox
  3. "Check here if an eDRS was sent.", eDRS_sent_checkbox

### DIFF
--- a/notes/csr.vbs
+++ b/notes/csr.vbs
@@ -539,46 +539,46 @@ End if
 '------Task/Step--------------------------------------------------------------Date completed---------------Notes-----------------------
 '
 '------Dialogs--------------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


courtrightd (approved):
Tested great, thank you!

Issue #2506, opened 2025-11-10
NOTES - CAF, fix bug reading EATS groups


MDR22:
User reported an issue where EATS groups were incorrectly populating in the dialog. Looks like script is not reading each line of the EATS groups properly

PR #2507, opened 2025-11-10
#2506 NOTES - CAF, Update EMReadScreen for the EATS groups

MDR22:
User reported bug where incorrect EATS groups added to dialog. Updated script to properly read each line of the EATS groups.


courtrightd (approved):
Looks like someone didn't quite finish their copy pasta. Thanks for fixing!

Issue #2509, opened 2025-11-12
On Demand bug happened with info format


C-Love:
Additional handling needed in the On Demand Waiver Applications script to ensure information read is a date and is reset on each loop. 

This does not appear to impact the run every time but it resolved the issue when it was triggered. Update as safety handling.

C-Love:
This is fixed.

PR #2510, opened 2025-11-12
#2509 - ODW Apps - PND2 date formats

C-Love:
Error reported by JH on 11/7 and this fix bypassed the issue though this issues hasn't come up again so is not a problem on every case. This handling should prevent this sporadic issue in the future.


courtrightd (approved):
excellent thanks!

PR #2528, opened 2025-12-02
Health Care eval access violation

courtrightd:
Script was checking for last_page_button constant (number >1000) instead of last_page constant, so potentially changing buttonpressed to undefined number/displaying a page that doesn't exist>
Also removed the display of ltc_intake_page when not required


C-Love (approved):
Looks good! Great catch.

Issue #2532, opened 2025-12-08
NOTES - CAF, improve handling for reading EI case note for JOBS panels with long employer names


MDR22:
Received report from worker regarding NOTES - CAF that was not pulling in details from EI case notes. After investigation, it looks like the script is not properly reading case notes for JOBS panels where the employer's name fills the entire line. Fix handling for these instances.

PR #2534, opened 2025-12-09
#2532 NOTES - CAF, Improve handling for reading employer names in case notes

MDR22:
Improved reading of employer names in earned income case notes to handle for employer names longer than 27 characters. I also fixed a small typo.


courtrightd (approved):
Thank you sir!

Issue #2547, opened 2025-12-26
Enrollment dates in MHC - Enrollment script need updating


C-Love:
Need to add 2026 dates

PR #2562, opened 2026-01-09
#2547 UCare HP Code Update for MHC Enrollment

C-Love:
UCare HP Code has updated as it was purchased by Medica.


courtrightd (approved):
Approved!

Issue #2491, opened 2025-11-04
ELIGIBILITY SUMMARY - GRH Supportive Housing now Automated in MAXIS


C-Love:
As of 11/1/2025 DHS has updated MAXIS to have the GRH Supportive Housing Disregard automatically included for cases that meet the requirement. Since we have been supporting a process to check these cases and stop the script run if it was not included correctly, we will be able to remove this functionality. 

One part of the functionality that should remain is the CASE/NOTE line when personal needs provides a better benefit than supportive housing disregard. This is still a FIAT process and the decision should be included in CASE NOTE. Information is available on SIR of these details. (https://www.dhssir.cty.dhs.state.mn.us/MAXIS/Lists/MAXIS%20Announcements/DispForm.aspx?ID=8879&ContentTypeId=0x01040097CAC0201808F844B3CD7E8E0C2ACAB600C88AE1669411024A9386CE6D01F7BD87 )

PR #2575, opened 2026-01-28
#2491 NOTES - Elig Summ- Ignore Hsg supp dsrgd when no income

C-Love:
Report in email from JD


courtrightd (approved):
Looks a-ok from here!

PR #2581, opened 2026-02-05
Eliminating 10 day variable from this script as it is already pulling from Func.

Farleight:
Also added email address


courtrightd (approved):
Approved!

Issue #2411, opened 2025-09-04
HC Evaluation - Ex Parte Review Dates Bug


courtrightd:
If the ER date is blank on REVW the script errors when attempting date logic.
Review script and update to not error and require entry of a date by the HSR before moving forward.
Make sure IR/AR logic is also functional.

PR #2617, opened 2026-04-28
#2411 HC Evaluation - Ex Parte Review Dates Bug

courtrightd:
#2411
Fixes a script crash that occured if the user had blanked out a date on the review popup - this is uncommon but can be done inadvertantly by the user during the script run while updating dates. A crash in this instance will lose their work and potentially get locked in maxis.

Also added data string to closing message when cases entered as not ex parte so we can pull the reason cases are failing.

### DIFF
--- a/notes/health-care-evaluation.vbs
+++ b/notes/health-care-evaluation.vbs
@@ -5980,24 +5980,25 @@ If HC_form_name = "No Form - Ex Parte Determination" Then
 			End If
 		End If
 
-		If mixed_household_notes <> "" Then
-			email_from = ""
-			email_recip = "ben.teskey@hennepin.us"
-			email_recip_CC = ""
-			email_recip_bcc = ""
-			email_subject = "Ex Parte Mixed Household Case - " & MAXIS_case_number
-			email_importance = 1
-			include_flag = False
-			email_flag_text = ""
-			email_flag_days = ""
-			email_flag_reminder = False
-			email_flag_reminder_days = ""
-			email_body = "This case has been indicated to have notes about a Mixed Household when evaluated in Ex Parte Phase 1." & vbCr & vbCr & "Notes:" & vbCr & mixed_household_notes
-			include_email_attachment = ""
-			email_attachment_array = ""
-			send_email = True
-			Call create_outlook_email(email_from, email_recip, email_recip_CC, email_recip_bcc, email_subject, email_importance, include_flag, email_flag_text, email_flag_days, email_flag_reminder, email_flag_reminder_days, email_body, include_email_attachment, email_attachment_array, send_email)
-		End If
+		'Pausing the mixed household emails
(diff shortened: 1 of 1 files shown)


notes/health-care-evaluation.vbs
```diff
@@ -6043,8 +6044,12 @@ If HC_form_name = "No Form - Ex Parte Determination" Then
 		CALL write_variable_in_case_note("---")
 		CALL write_variable_in_case_note(worker_signature)
 
-		'Script end procedure
-		script_end_procedure("Success! The ex parte review information has been added to the CASE NOTE")
+		'Script end procedure. Adds denial reason for non passed cases for data analysis purposes.
+		If ex_parte_determination = "Cannot be Processed as Ex Parte" Then
+			script_end_procedure("Success! The ex parte review information has been added to the CASE NOTE. Case cannot be processed as ex parte due to: " & trim(replace(ex_parte_denial_select, "Select or Enter Reason for NOT Ex Parte", "")))
```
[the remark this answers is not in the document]
courtrightd: Leaving this as is - the suggested change would make it harder to extract the preset options from the string, and all cases without a preset will be categorized as "other" for data purposes.

notes/health-care-evaluation.vbs
```diff
@@ -5689,62 +5689,54 @@ If HC_form_name = "No Form - Ex Parte Determination" Then
...
+							cancel_confirmation
+
+							'TODO - read DISA and display waiver
+
+							EMReadScreen stat_check, 4, 20, 21
+							EMReadScreen revw_panel_check, 4, 2, 46
+							EMReadScreen hc_revw_pop_up_check, 20, 4, 32
+
+							If hc_revw_pop_up_check <> "HEALTH CARE RENEWALS" Then
+								If hc_revw_pop_up_check = "REVW" Then
```
C-Love: This should be the variable 'revw_panel_check'

notes/health-care-evaluation.vbs
```diff
@@ -5783,44 +5775,50 @@ If HC_form_name = "No Form - Ex Parte Determination" Then
...
```
C-Love: Should be 'revw_panel_check'

C-Love (commented):
There is a lot of duplication between mixed household and not.

C-Love:
The revw_panel_check variable issue is not new and basically just means the If statements on line 5790 and 5704 will never be 'true'. This is an old error and doesn't seem to be impeding functionality in any way. Basically, it just falls to the next line 'ElseIf stat_check = "STAT" Then' and triggers the navigation there.

C-Love (approved):
This should work fine. It would be best to refactor and have the dialog only defined and displayed once, using a 'If mixed_household' within the error handling and HCMI review to correctly evaluate but not needing to recreate the entire loop twice.

PR #2619, opened 2026-04-30
#2618 NOTES - LTC-5181, Fix OS ADDR update details in CASE/NOTE

MDR22:
Script incorrectly wrote variables to CASE/NOTE for ADDR updates, fixed the variables to write correctly.


C-Love (changes requested):
The updates are good. However, it looks like the script (at least when run by OS staff) will always delete phone, email, and mailing address information from ADDR. This is probably not what we want to have happen.

C-Love:
We could probably fix this with a different pull request. There's nothing in this pull request causing the issue.

Issue #2616, opened 2026-04-27
NOTES - LTC-5181, improve ADDR update functionality


MDR22:
Received feedback that ADDR update functionality for OS staff completely clears ADDR panel when making updates. Requested that ADDR updates only overwrite where changes are needed, not the entire panel. Also requested that dialog include fields to update mailing address.

PR #2630, opened 2026-05-14
#2616 NOTES - LTC-5181,  Add mailing address update functionality, improve ADDR update functionality

MDR22:
Added functionality to update mailing address, improve ADDR update functionality so that it does not overwrite ADDR panel.


courtrightd (approved):
Approved!

Issue #2634, opened 2026-05-20
NOTES - LTC-5181, remove footer month, don't update VER or living situation


MDR22:
Received request from OS staff to remove the footer month selection on initial dialog, and to NOT update the VER or living situation - leave as is on the ADDR panel

PR #2635, opened 2026-05-20
#2634 NOTES - LTC-5181, Remove footer month/year from first dialog, remove ver and livi…

MDR22:
…ng sit from case note and ADDR updates, fix bug with footer month/year for some ADDR updates


courtrightd (approved):
Approved!

